PR #4, opened 2026-07-14
docs: stamp the live map as latest-<hash>, back-fill a v0.1.0 parity map

calvinchengx:
Two requested changes to the parity versioning.

## 1. `latest-<hash>` stamp
`git describe` produced **`v0.2.0-2-gb1e3520`** for the live map — which reads like a release that doesn't exist. The live map now stamps as **`latest-b1e3520`**: plainly the moving tip of `main`.

Sitting exactly on a release tag still stamps that tag (**`v0.2.0`**), so the release moment stays unambiguous. Verified both paths. The now-redundant `(unreleased)` suffixes are dropped.

## 2. v0.1.0 back-filled (reconstructed)

**Why it was missing:** `v0.1.0` shipped *before* the parity map existed, so there was nothing to show — and a diff can't invent it, because the map is the *input* to the diff, not the output.

**Why it can't be derived:** the map is a hand-authored judgement (🟢 Real vs 🟡 Emulated), not something inferable from a code diff. And the tag is immutable, so the map can't be added retroactively to it.

**So:** `collectParity` now falls back to a hand-authored back-fill at `docs/parity-snapshots/<tag>.md`, flagged `reconstructed` — the page carries a **caution banner**, and the dropdown/index say "(reconstructed)". This is a retrospective reading, **not** a document published at the time.

`docs/parity-snapshots/v0.1.0.md` is reconstructed from that tag's evidence:
- `docs/13-roadmap.md` — 28 `[x]` done, 12 `[ ]` open
- `docs/12-e2e-matrix.md` — what CI actually proved
- the v0.1.0 source tree — a 🔴 usually means the package is simply **absent** (`internal/tds`, `internal/pipeline`, `internal/warehouse`, `onelake/blob.go`, `shortcuts.go`, `python/notebookutils`)

Row names match the current map, so the changelog now shows the real v0.2.0 story — **v0.1.0 → v0.2.0**: Blob surface, Delta put-if-absent atomicity, shortcuts, Livy (+HC), notebookutils, the warehouse/TDS stack and the pipeline interpreter all **🔴 → 🟢**; client witnesses go from **one** (`fabric-cicd`) to **eleven**.

The **Scope boundary** section is deliberately *not* reconstructed — that's later editorial analysis, not a capability claim, and inventing it for v0.1.0 would be dishonest.

> Please sanity-check the status calls in `docs/parity-snapshots/v0.1.0.md` — they're my reading of the evidence, and you'd know better where I've mis-graded a row.

Verified: full Astro build (25 pages) + browser render of the v0.1.0 snapshot, dropdown, and changelog.


calvinchengx:
Added a third change to this PR: **the version picker now also lives in the right sidebar**, next to "On this page" (the in-page one scrolls away).

- `sync-docs` additionally emits `src/data/parity-versions.json` (generated, gitignored) — an `.astro` component can't shell out to git, so it reads the manifest.
- **Route-scoped:** renders only on pages it can switch between (the live map + each snapshot). Not on other docs, and not on the changelog — that spans all versions rather than being one.
- **Overrides `TableOfContents`, not `PageSidebar`** — deliberately. Starlight renders the ToC inside a width-constrained `.sl-container` whose styles are component-scoped. A picker hung off `PageSidebar` sits *outside* it and blows out to **1452px, overflowing the viewport** (I measured it). Overriding `TableOfContents` puts it inside that container, inheriting the sidebar width (268px) rather than duplicating Starlight's own "MAGIC NUMBER 🥲" width calc.
- Desktop only by construction (mobile uses `MobileTableOfContents`); narrow screens still get the in-page picker.

Verified in a browser at 1500px: fits the container exactly, no overflow, and selecting "v0.1.0 (reconstructed)" from the **sidebar** navigates to that snapshot with the right option pre-selected.

calvinchengx:
Updated per feedback — **one picker, in the top nav**, and the "reconstructed" labelling is gone.

**Picker moved to the top nav.** The in-page picker is removed; the sole picker now sits beside the theme select, visible from anywhere on the page. Still route-scoped: it renders only on pages it can switch between (live map + each snapshot), so every other page gets the stock header. This also replaces the right-sidebar override, which is removed.

Implemented by overriding `ThemeSelect` rather than `Header` — Starlight's `Header` has no slot, and copying `Header.astro` wholesale would fork its layout and drift on upgrade.

> **Caveat worth knowing:** Starlight's header right-group is `sl-hidden md:sl-flex`, so the picker (like the theme select itself) appears at **≥50rem / 800px**. Below that it's hidden — `MobileTableOfContents` isn't overridable, so there's no clean way to reach phones. Narrow screens still get every version via the **Parity history** index page. Better than the sidebar version, which needed ≥1152px.

**"reconstructed" removed** — the `(reconstructed)` option suffix and the amber caution banner are both gone. The dropdown now reads a clean `v0.1.0`.

One thing I kept, and want to flag rather than slip past you: the v0.1.0 page still **opens by saying it was written after v0.1.0 shipped** — but as ordinary prose, not a banner. The docs are public, and without that line a reader would reasonably assume the map was published at v0.1.0. If you'd rather it go entirely, say the word and I'll strip it.

calvinchengx:
Sidebar IA updated: **parity now has its own section, with the live map first.**

Before, the map sat under **Project** carrying its full title ("Feature parity: fabric-emulator vs. real Microsoft Fabric") while its history and changelog lived in a separate "Parity history" group — three related pages split across two sections.

Now:

```
Parity
  Parity            -> /fabric-emulator/17-parity/      (the live map)
  Parity history    -> /fabric-emulator/parity-history/
  Parity changelog  -> /fabric-emulator/parity-history/changelog/
```

- "Parity history" group renamed to **Parity**.
- The live map moved out of **Project** to be its **first** entry, labelled just **Parity** (the long title still heads the page itself).

**One judgement call to confirm:** I read "named simply as `parity`" as the *sidebar label*, so the map's **URL is unchanged** (`/17-parity/`) and no published link breaks. If you actually meant the slug — `/fabric-emulator/parity/` — I can do that too, but it needs a bit more: a redirect from `/17-parity/` (there's precedent in `astro.config`), a tweak to sync-docs' `NN-*.md` filter, and updating intra-doc links. Say the word.

calvinchengx:
**Parity map de-numbered:** `docs/17-parity.md` → `docs/parity.md`, published at **`/fabric-emulator/parity/`**. It's a living reference, not a chapter in the reading order.

- **Old URL redirects** (`/17-parity/` → `/parity/`), alongside the existing renumbering redirects — nothing published breaks.
- `sync-docs`: the `/docs` scan and the intra-doc link rewriter now accept the un-numbered map as well as `NN-name.md` chapters.
- `PARITY_RE` matches `docs/parity.md` **and** `docs/17-parity.md` — older tags still carry it numbered, so **v0.2.0's snapshot still resolves**. The `[/-]` boundary keeps it from matching e.g. `disparity.md`.
- Sidebar slug → `parity`; the "See parity.md" link in `18-semantic-model-references` follows the rename.

Verified: `/parity/` 200, `/17-parity/` → redirects → 200, both snapshots still build, and **zero** stale `/17-parity/` links anywhere but the redirect stub.

### A latent bug this surfaced (fixed here)
`collectParity` read the live map via `git show HEAD:`, while the rest of sync-docs renders `/docs` **from disk**. Those disagree whenever the doc has uncommitted edits — the page built at `/parity/` while the picker still pointed at `/17-parity/`, so the route match failed and **the picker silently hid itself**. Invisible in CI (clean checkout = HEAD), but it would bite anyone editing the parity map under `astro dev`. The live map now comes from the working tree; tags still come from git.

### Note
This leaves a **gap at 17** in the numbered sequence. I left 18/19 alone deliberately — renumbering them would break their published URLs for no real gain. Say the word if you'd rather close the gap.

PR #22, opened 2026-08-06
events: generate the portal contract as TypeScript

calvinchengx:
Closes the last gap in the SSE contract work. `AllKinds` already guaranteed the portal *subscribes* to every kind (#19) and the wire tests proved every kind is framed and filterable (#21). Neither said the portal can **render** one.

### What changes

`scripts/gen_event_kinds.py` now emits `portal/src/eventKinds.ts` instead of `.js`:

- `EVENT_KINDS`/`VIEW_KINDS` as `as const`, so `EventKind` and `ViewKind` are unions derived from the same lists;
- `EmulatorEvent` — the `store.Event` struct as an interface, Go comments carried through, so a renamed field is a compile error instead of a silent `undefined`. Named `EmulatorEvent` because `Event` is a DOM global that would shadow the type every event handler is checked against;
- `isEventKind` / `isViewKind` guards, because `/_emulator/events` is a public stream that anything may point at a different build.

An unmapped Go type stops the generator rather than becoming an `any` nobody notices.

`Flow.svelte` is `lang="ts"`. `describe()`'s switch is exhaustive over `ViewKind` and `ingest()` has a `never` arm for a kind that is neither `dropped` nor renderable — so both ways of adding a kind in Go break the build.

Types stop at the generated contract. The lineage/table/workspace endpoints stay `any`, with a comment saying why: a hand-written interface for an ungenerated endpoint is an unchecked claim about the server, which is the exact drift that made a hand-written kind list a bug.

### What proves it

`scripts/check_kind_exhaustiveness.py` declares a kind nothing handles and requires the portal to stop compiling — once in `AllKinds` + `ViewKinds` (caught by `describe()`), once in `AllKinds` only (caught by `ingest()`). Each mutation is bracketed by a clean run, so a check that fails for an unrelated reason cannot pass for the right one. Verified non-vacuous: removing both guards makes it report FAIL twice.

New CI job `portal-types`, separate from `portal` the way `ruff + ty` is separate from the Go build — a red X means the types are wrong, not the bundle is broken. Errors only; the warnings are svelte-check's CSS parser not knowing Tailwind's `@apply` plus unused-selector notes that predate this job.

### Two incidental fixes the type-checker found

- `button.svelte` destructured `class`/`disabled`/`children` with no default, which infers them as **required** — every call site owed a `class` it has never needed.
- The semantic-model node carried `transform` on an `<a>`, which inside an `<svg>` is still typed as the HTML anchor. Moved to a wrapping `<g>`, where a group's placement belongs. That branch had **no test** — added one, and confirmed it fails when the transform moves.

### Verified locally

`go build`/`vet`, `store` + `server` tests, `make check`, `make portal-types`, 86 portal unit tests (up 1), `pnpm build` reproducible against committed `portal/dist`.


calvinchengx:
Pushed `25083ea`: the three `notebookutils` legs were mine, and the diagnosis is a good one.

`[tool.coverage.run] source = ["scripts", …]`, so adding ~120 statements of untested Python — the enlarged `gen_event_kinds.py` plus the new `check_kind_exhaustiveness.py` — took the total from above the floor to **63%**, and `--cov-fail-under=70` did its job.

The fix is tests rather than an `omit` entry, because the generator is a **parser** and it is the piece everything downstream trusts. A parser that misreads its input does not fail; it emits something plausible — an unrecognised Go type becomes `any`, a missed `,omitempty` becomes a required field the server never sends, and both compile.

- `python/tests/test_gen_event_kinds.py` — 20 cases: the union derives from the same array the runtime list does, `dropped` is carried but never a filter, labels come from the Go comments, `omitempty` → optional, pointer → optional number, `kind` → `EventKind` not `string`, `Event` → `EmulatorEvent` (a DOM global would otherwise be shadowed), field comments carried across; plus six malformed inputs that must **stop** it rather than guess. The happy path runs against the real `bus.go`, so a fixture cannot keep passing after the struct changes shape.
- `python/tests/test_check_kind_exhaustiveness.py` — the mutation harness's weak point is not the compiler, it is the regex: one that silently matches nothing leaves the tree untouched, the build passes because nothing was ever wrong, and the harness reports success. So: the mutation lands, the two mutations genuinely differ, the mutated Go is still readable by the generator (one it choked on would fail for the wrong reason), and a bus.go it cannot edit stops it loudly.

Coverage 63% → **75%**. `gen_event_kinds.py` is at 99%; `check_kind_exhaustiveness.py` at 38%, the remainder being the orchestration that shells out to svelte-check — which is the CI job's job, not a unit test's.

PR #25, opened 2026-08-06
portal: all TypeScript, 100% coverage, and a coverage badge

calvinchengx:
Two questions, both answered by doing the work: what was still JavaScript, and what the portal's test coverage actually was.

## The portal had never been measured

Go has a 90% floor and Python a 70% one. The client — the part a user actually looks at — had **no coverage measurement at all**. First reading: **86.4%** of our own statements, with `App.svelte` (92 statements) and `Identities.svelte` (37) at **zero**.

Now:

```
Statements   100%    (1486/1486)
Functions    100%    (339/339)
Lines        100%    (741/741)
Branches     92.51%  (470/508)
```

203 tests, up from 85. Gated in `portal/vite.config.ts` at exactly those numbers, and CI runs the suite with `--coverage` so the thresholds actually apply.

**Why branches stop at 92.51%, and why that is the honest ceiling.** Svelte wraps every `{interpolation}` in a nullish guard, so `{j.status.toLowerCase()}` compiles to a two-armed branch whose second arm needs `status` to be nullish — which would have thrown inside `toLowerCase()` first. 38 of the 508 arms are that shape, every one on a markup line with no source-level conditional. I chased the real ones down to zero: the remainder is the compiler's, not the suite's. `--coverage.experimentalAstAwareRemapping` moves it by 0.8 points, so it is not a reporting-mode problem either.

Exactly **two** statements carry `/* v8 ignore next */`, both defensive guards no input can reach — a second `connect()`, and `edgePath` on a link whose node is gone — each with the reason written beside it. I left the guards in; deleting correct defensive code to satisfy a counter is the wrong trade.

## Two real bugs fell out of it

- **`ModelDetail` crashed on a table with no `measures` key.** `sampleQuery` guarded `t.measures || []`; the markup three lines down read `t.measures.length` directly. Same for `t.columns` and `model.relationships`. The page threw while the query box derived from the same model perfectly happily. Hardened to match the intent already in the file.
- **The e2e badge overstated the fleet by three.** `coverage_badges.py` counted every top-level CI job that was not on a five-name exclusion list, so `lint`, `coverage-e2e` and my own `portal-types` were being published as end-to-end suites. 38 → **35**.

## Everything is TypeScript

| | Before | After |
|---|---|---|
| Components | 14 `lang="js"`, 1 ts | **15 `lang="ts"`** |
| Modules | `api.js`, `router.js`, `main.js` | **`.ts`, typed** |
| Tests | 13 `.js` | **19 `.ts`** |
| Build config | vite / vitest-setup / playwright `.js` | **`.ts`** |
| `src/lib/utils`, 6 barrels | `.js` | **`.ts`** |

`svelte-check`: **0 errors over 718 files** under `strict`.

The line I held: **API responses stay `any`**, with the reason in `api.ts`. A hand-written interface for an ungenerated endpoint is an unchecked claim about the server — precisely the drift that made a hand-written event-kind list a bug. The event stream is typed to the hilt because its contract is *generated* from `internal/store/bus.go` and cannot disagree.

`src/testing.ts` replaces 19 hand-rolled `Response` literals with one cast that carries its reason, plus the shared `FakeEventSource` that four files had each declared for themselves.

## Left deliberately

The **23 vendored shadcn components** under `src/lib/components/ui/` are still `lang="js"`. They are regenerated by `shadcn-svelte add`, so typing them forks upstream and the next update overwrites it. They are excluded from coverage for the same reason, with the reason in the config rather than silent. Say the word and I will convert them.

## The badge

`[![portal coverage]](…/coverage-portal.json)` beside the Go and Python ones, same self-hosted shields `endpoint` document, written by the same script — a second publisher would be a second place for the numbers to disagree. The badge step gained Node so all five numbers still come from one place.

## Verified

`go build`/`vet`, `make check`, `ruff`, `ty`, 152 Python tests, 203 portal tests, `svelte-check` clean, `vite build` reproducible against the committed `portal/dist`.


calvinchengx:
**Correction, and the vendored components are now TypeScript too.**

I said typing them would fork upstream. That was wrong: shadcn-svelte emits TypeScript **by default** — JavaScript is the opt-out — and this repo's \`portal/components.json\` simply carried \`"typescript": false\`. It was a setting, not an upstream constraint.

So I flipped it and re-added the components through the CLI rather than hand-converting them:

\`\`\`
pnpm dlx shadcn-svelte@latest add badge button card checkbox separator table --overwrite
\`\`\`

All 19 vendored components are now \`lang="ts"\`, and their six barrels are \`.ts\`. **`find` now reports no `.js` anywhere in `portal/` outside `node_modules`, `dist` and coverage output.**

Three things worth noting:

- **The local fork went away.** My earlier `button.svelte` patch — adding `= undefined` to `class`/`disabled`/`children` so they were not inferred as required — is gone, because the TypeScript variant states optionality in `ButtonProps`. The JS variant could not express that, which is exactly why the patch existed.
- **`utils.ts` needed upstream's type helpers.** The TS components import `WithElementRef`, `WithoutChild`, `WithoutChildren` and `WithoutChildrenOrChild` from `$lib/utils`, which the JavaScript variant has no use for and therefore never generated. Added, with a note saying why they appeared.
- **One dependency arrived:** `@internationalized/date`, a `bits-ui` peer that had been unmet. Nothing imports it directly; it satisfies the peer range.

`tsconfig.json` loses `allowJs`/`checkJs` — there is no JavaScript left for them to admit.

Re-verified end to end: `svelte-check` 0 errors over 718 files, 203 tests, coverage still 100% statements / functions / lines, `vite build` reproducible against the committed `portal/dist`, and the Playwright smoke test mounts the rebuilt bundle in a real browser.

PR #34, opened 2026-08-06
docs: size the stack by what you run, not what you start

calvinchengx:
Measured resource guidance in `docs/27` and `README.md`, so a user can pick a mode without booting all of them to find out.

## The finding that reshaped the advice

**Footprint is driven by the workload, not by the service list.** Measured on the same container:

| | Just booted | Running PySpark |
|---|---|---|
| `sail` | **36 MB** | **~1.9 GB** |
| `fabric-emulator` | **11 MB** | **500 MB – 1 GB** |

A 50× spread. So *starting an engine you do not drive is nearly free*, and the number to size against is the working one — which the docs previously conflated with idle.

## What the docs now say

`docs/27` leads with a **choose-by-what-you-need** table rather than by mode name:

| I want to work on… | Command | Services | Idle | Give the runtime |
|---|---|---|---|---|
| control plane, OneLake, RBAC, git, CI/CD | `make up-lite` | 3 | ~65 MB | 2 GB |
| …plus Spark and the T-SQL warehouse | `docker compose up` | 6 | ~1 GB | 8 GB |
| …plus catalog, lineage, glossary | `make up` | 11 | ~3 GB | 12 GB |
| …plus KQL | `--profile rti` | 12 | +4 GB | +4 GB |
| …plus a shell in the Flow view | `--profile terminal` | 13 | +5 MB | +5 MB |

Then a per-service table with three columns — just booted, light work (delta-rs write + two Copy activities), heavy work (real PySpark medallion + catalog ingest) — and the three things that actually decide the setting.

**CPU is documented for the first time, and it is not the constraint:** peak during light work was `sqlserver` 18%, `sail` 12%, `fabric-emulator` 1.2% — fractions of a *single* core. Four cores is ample, six to eight for PySpark or warehouse work. Noted that `sqlserver` is the one amd64 container, so Apple silicon pays Rosetta for it.

## Corrections to my own earlier table

The version I landed in #29 mixed an idle stack with a loaded one and had no CPU at all. Specifically wrong: `sqlserver` "380 MB idle" — it is **881 MB**, because SQL Server reserves at startup; the 380 came from an instance that had been idle long enough to trim. `sail` and `fabric-emulator` were listed as "small" with no figure.

`README.md` and `docs/01-quickstart.md` now agree with it — the quickstart's "8 GB" says what it covers, and that the same six services idle at about 1 GB.

## Provenance

14-core / 36 GB host, OrbStack, on the images this repo pins. Idle and light-work figures come from a stack booted for the purpose on an isolated project and torn down after; heavy-work figures from a full medallion run. Every number was read from `docker stats`, not estimated.

Docs only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,19 +171,19 @@ identities), so it could equally point at a real Entra tenant.
 | `-f docker-compose.spark-jvm.yml` | **swaps** Sail for JVM Spark, buying the RDD API, structured streaming, `OPTIMIZE`/`VACUUM` and Java/Scala UDFs at the cost of image size ([docs/20](docs/20-lakesail-engine.md)) |
 
 Profiles pull nothing unless asked for — but **`make up` asks for `governance`
-on your behalf**, so it starts 11 services rather than 6. `make up PROFILE=`
+on your behalf**, so it starts 12 services rather than 6. `make up PROFILE=`
 gives the lean stack.
 
 **How much machine you need.** Give the container runtime **8 GB** for the
-default six, **12 GB** with governance, **2 GB** for the lite pair, **16 GB** for
-everything at once. Four cores is ample; six to eight if you run PySpark or
-warehouse queries.
+default six, **13 GB** with governance and Airflow, **2 GB** for the lite pair,
+**17 GB** for everything at once. Four cores is ample; six to eight if you run
+PySpark or warehouse queries.
 
 Those numbers are for *working*, and idle is nowhere near them — a freshly booted
 lite stack is 65 MB, and the whole default six is about 1 GB. The spread is the
 work, not the container count: Sail costs 36 MB to start and ~1.9 GB to run
 PySpark through. So starting an engine you do not drive is nearly free, and
-[the per-service measurements](docs/27-running-modes.md#what-it-actually-costs)
+[the per-service measurements](docs/27-running-modes.md#what-it-costs-to-run)
 are what to size against.
 
 ## Getting started on Linux, macOS or Windows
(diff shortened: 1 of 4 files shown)


calvinchengx:
Rebased onto `main` and **substantially reduced**, because the collision I flagged happened: the Airflow change landed its own `## What it costs to run` section covering the same ground, and better in places — it measured Airflow (1.09 GiB) and image sizes, which I had not.

Shipping two overlapping resource sections would be worse than either, so this PR now adds only what that section lacks:

1. **A capability-first chooser** at the top — "I want to work on the control plane…" → command, service count, idle, give-the-runtime. The existing table is organised by mode name, which requires already knowing what a mode contains.
2. **CPU, which the page has never stated** — peak during light work was `sqlserver` 18%, `sail` 12%, `fabric-emulator` 1.2%, all fractions of a *single* core. Four cores ample, six to eight for PySpark. Plus the Apple-silicon caveat: `sqlserver` is the only amd64 container, so Rosetta makes it cost more there.
3. **A reconciled `sqlserver` figure.** Their table says 380 MiB idle, I measured 880 MB. Both are right — 880 is a fresh boot, 380 is the same instance after it has trimmed. It now says so, because 880 is what a reader will actually see.

I deleted my own duplicate section (44 lines) rather than keep the better-sounding one.

### Two things fixed in passing

**`main` is failing `make check`.** `docs/39-run-multiple-parity-plan.md` is published by the site but in no sidebar group, so `check_docs_sidebar` fails — `40-rest-connector-plan` was added to `website/astro.config.mjs` and 39 was missed. Added; the gate now reports *41 published docs, all reachable*, and the site builds 71 pages.

**`README.md` disagreed with `docs/27`.** The Airflow change updated the doc to 12 services / 13 GB / 17 GB but left the README saying 11 / 12 / 16. Both, and the quickstart, now agree — verified against `PROFILE ?= --profile governance --profile airflow`, which resolves to exactly 12 services and 14 with everything.

`make check` clean, docs site builds, no dangling anchors.

PR #44, opened 2026-08-07
Add infix operators to the DAX evaluator

### LINKED ISSUE #42: DAX evaluator rejects all infix operators (+, -, *, /, &, comparisons)
## Summary

The DAX evaluator's lexer accepts no infix operators at all. `+`, `-`, `*`, `/`, `&` and comparisons each fail at tokenisation with `unexpected character "<c>" in DAX`, both when written inline in a query and when they appear inside a stored measure's expression.

Function calls are unaffected, so `DIVIDE([A], [B])` works while `[A] / [B]` does not.

## Reproduction

Against a published import model, over `executeQueries` (emulator build `v0.17.0-0ba3d8b`):

```
EVALUATE SUMMARIZECOLUMNS(Reporting[FiscalQuarterLabel], "X", [Revenue USD])
  -> 200, {"[X]": 129303176.01}

EVALUATE SUMMARIZECOLUMNS(Reporting[FiscalQuarterLabel], "X", [Revenue USD] + [Cancelled Revenue])
  -> 400 {"code":"DAXQueryError","message":"unexpected character \"+\" in DAX"}
```

Same failure for `-`, `*`, `/`, `&`, and `>` (inside `IF`). `DIVIDE([Revenue USD], [Units Sold])` returns 200.

## Why it is easy to miss

A measure whose expression uses an operator **publishes without complaint** and only fails when something queries it. A model can therefore pass creation, appear complete in the portal's model detail page, and still be unqueryable in part. In `contoso-data-platform` the published model carries

```
measure 'Gross Revenue' = [Revenue USD] + [Cancelled Revenue]
```

which was green for weeks because nothing had asked for that particular measure. It surfaced only when a reconciliation check queried each measure in turn.

## Where

`internal/semanticmodel/dax.go`, the lexer around line 60. The `switch` handles whitespace, quoted/bracketed names, `(`, `)`, `,`, numeric literals and alpha identifiers; everything else hits the `default` branch and errors. Note `-` is accepted only as the sign of a numeric literal, so subtraction between measures fails on the same path.

## Suggested direction

Two pieces, in order:

1. Lex the operator characters as tokens (`+ - * / & = <> < <= > >=`).
2. Parse them with precedence — multiplicative over additive over comparison over `&` — rather than extending the current call-shaped parser, since operands can themselves be measure references, column references or nested calls.

Worth pairing with a parity witness that a measure defined with an operator round-trips: publish, query, and get the arithmetic result. That is the check that would have caught this.

## Impact

Any consumer that writes ordinary DAX hits it immediately — measure arithmetic is one of the most common things a semantic model contains, and real Power BI supports it. It also blocks reconciling a model measure-by-measure against the warehouse, which is the check that proves published numbers are correct rather than merely present.

calvinchengx:
Fixes #42.

## The bug

The DAX lexer had no case for any operator character, so `+ - * / &` and every comparison failed at tokenisation — `DIVIDE([A], [B])` worked while `[A] / [B]` did not. `-` was accepted only as the sign of a numeric literal, so subtraction between measures failed on the same path.

## The fix

- **Lexer**: `+ - * / & = <> < <= > >=` become `tOp` tokens. `-` is now always an operator; folding it into number literals would lex `[A] -1` as two operands and leave the parser no subtraction to see, so negation moved to the parser.
- **Parser**: precedence climbing over DAX's own order — unary `-`, then `* /`, then `+ -`, then `&`, then the comparisons — plus parenthesised sub-expressions and string literals (`&` needs operands).
- **Evaluator**: BLANK coerces to 0 in arithmetic and to `""` in concatenation; comparisons yield TRUE/FALSE; non-numeric text errors rather than silently becoming 0. `a / 0` errors and points at `DIVIDE` — real DAX yields Infinity, which has no JSON encoding, so inventing a number would be worse than refusing.
- **`IF(cond, then[, else])`**: the reported repro put the failing `>` inside `IF`, and a comparison is not much use without a consumer. Omitted else → blank.
- **Measure tails are refused**: a measure expression is parsed on use and the parser used to stop at the first token it did not understand *without complaining*, so a half-understood measure evaluated to the half it read. That now errors.

## The witness

A measure using an operator **publishes without complaint** — the definition is stored verbatim and nothing reads the DAX until a query names the measure. A model can be created, listed, and shown complete in the portal while part of it is unqueryable. Creation is no evidence; only a query is.

So `retail.bim` gains two measures defined with operators (`Units Delta`, and `Units Growth Pct` which nests it), and:

- `TestExecuteQueriesEveryPublishedMeasureAnswers` publishes the model, then queries **every** measure in it and asserts the arithmetic — the reconciliation check that would have caught this. It enumerates from the fixture, so a new measure is covered automatically.
- Two golden queries (`Operator Measure Asset`, `Inline Operator Precedence`) run in the Go golden test *and* the Python `executeQueries` e2e. The precedence one pins `TY + LY*2` against `(TY+LY)*2` — different values under either reading.
- `dax_operators_test.go` covers precedence, associativity, unary/parens, BLANK coercion, concatenation formatting, and the error paths.

Verified non-vacuous: with `dax.go` reverted, the new tests fail with the reported `unexpected character "+" in DAX`, and the witness fails on exactly the two operator measures while the other four still answer.

## What consumer-side probing added (commits 2–4)

Independent probing against a real Contoso model — see the evidence comment below — turned up a defect worse than #42, reached by following this PR's own advice. The three follow-on commits are fallout from the first, not new ground:

- **c3b3d29** — the bare-column error said `used outside an aggregation`, which reads as "`&` is broken" to someone building a label. Reworded. The behaviour is correct and unchanged: real DAX refuses a bare column in a `SUMMARIZECOLUMNS` extension expression too, because a filter context supplies no row to read one value from.
- **b31a5f8** — **`SUM` over a text column returned `0`.** Not an error, a number: `"FY" & SUM(Reporting[FiscalYearLabel])` produced a label reading `FY0`, and `SUM(text) = 0` evaluates true, so a downstream check passes on it. c3b3d29 made this worse by recommending `SUM(...)` for the exact case where it lies. `SUM` now refuses non-numeric values at the source, and the hint is type-aware — group-by leads for every column, `SUM` is offered only when the column's dataType can be summed, and an unknown dataType stays silent. `DIVIDE` had the identical coercion and this PR is what made it reachable (string literals were not valid scalars before), so it is fixed here too.
- **75e40cf** — b31a5f8 then refused `""`, contradicting its own rule that blank contributes nothing: `""` is what a CSV or warehouse export writes for a missing number, so a text-encoded numeric column failed on its first gap. Empty and whitespace-only *column values* are now blank; a literal `""` in an expression still errors, since that is deliberate text. Refused values are quoted, so the empty case reads `the value "" is not a number` rather than trailing into a hole. The numeric-text boundary is now a decision with a test walking it: padding is tolerated as an export artefact, while `1,234`, `$99` and `50%` refuse, because guessing a locale is how a total silently loses three orders of magnitude.

Kept in this PR rather than split out: the defect is only reachable through the operator work, and the message that steered people into it is this PR's own.

`go build`, `go vet`, `go test ./...` and `make check` all pass.


calvinchengx:
## Consumer-side verification against a real model

The fixtures in this PR are hand-computed by me, so on their own they can only catch me disagreeing with myself. The runs below come from a parallel session working on `contoso-data-platform`, which built images from this branch and probed them against a **real Contoso model** with a freshly provisioned platform (16/16) behind each run. Posted with their permission; the findings and the numbers are theirs.

### 1. Operators, at `b1b017a`

Query shape `EVALUATE SUMMARIZECOLUMNS(Reporting[FiscalQuarterLabel], "X", <expr>)`, build `vpr44-b1b017a`:

```
OK    [Gross Revenue]  (measure defined as [A] + [B])
        FY27 Q2=132101515.29, FY27 Q1=40146.78
OK    inline [Revenue USD] + [Cancelled Revenue]
        FY27 Q2=132101515.29, FY27 Q1=40146.78
OK    inline [Revenue USD] * 2
        FY27 Q2=258606352.02, FY27 Q1=75963.32
OK    IF([Revenue USD] > 0, 1, 0)   (the issue's repro)
        FY27 Q2=1, FY27 Q1=1
OK    [Carried Rate Share]  (DIVIDE, unchanged)
        FY27 Q2=0.2848370567259045, FY27 Q1=0
```

Three things here matter more than the passes:

- **Both quarters reconcile**, not just the large one. Q2: 129,303,176.01 + 2,798,339.28 = 132,101,515.29. Q1: 37,981.66 + 2,165.12 = 40,146.78. The magnitudes are four orders apart, so a coincidence would not survive both.
- **Inline and stored-measure forms agree exactly**, which is the actual claim of #42 — the operator behaves the same whether it arrives in the query string or out of the model definition. This is the half a lexer-only fix would have left broken, since `IF` was still unsupported at that point.
- **`DIVIDE` is unchanged**, the regression check.

### 2. Two-surface reconciliation

`contoso-data-platform`'s reconciliation gate, DAX through `executeQueries` against a plain `GROUP BY` on the warehouse over TDS — no shared code path between the two:

```
FY27 Q1  revenue_usd            model=37,981.66       warehouse=37,981.66       ok
FY27 Q2  revenue_usd            model=129,303,176.01  warehouse=129,303,176.01  ok
RECONCILED 2 fiscal quarter(s), two surfaces agree — total revenue 129,341,157.67
```

### 3. How the `FY0` defect surfaced, at `c3b3d29`

The reworded error rendered correctly against a real column, and its second suggestion (group by it) returned the right full-year total. Its *first* suggestion, followed literally on a text column, did not:

```
"FY" & SUM(Reporting[FiscalYearLabel])   ->  OK   "[X]": "FY0"
```

No error. A label reading `FY0`, from following the advice the message gave. Root cause was broader than the message:

```
SUM(Reporting[FiscalYearLabel])        ->  OK   0        <- text column summed to zero
SUM(Reporting[FiscalYearLabel]) = 0    ->  OK   true
```

### 4. After the fix, at `b31a5f8`

No regression — the platform came up 16/16 and the reconciliation gate still agrees on 129,341,157.67, so no Contoso measure had been standing on a coerced text sum. Real text columns now refuse and name the offender:

```
SUM(Customer[CustomerId])       cannot sum Customer[CustomerId]: the value C-0004649 is not a number
SUM(Revenue[OrderDate])         cannot sum Revenue[OrderDate]: the value 2026-07-22 is not a number
SUM(Country[Country])           cannot sum Country[Country]: the value SG is not a number
"FY" & SUM(...FiscalYearLabel)  cannot sum Reporting[FiscalYearLabel]: the value FY27 is not a number   <- the FY0 path, closed
SUM(Reporting[Units])           OK 1,280,817
```

Contoso has no numeric-looking text, so they published a throwaway `NumericTextProbe` model with one column per shape — one column each, because a single grouped query aborts on the first failing group and would only ever report one shape:

```
'123'      OK 123        '1,234'  FAIL ...the value 1,234 is not a number
'00123'    OK 123        '$99'    FAIL
'1.5'      OK 1.5        '50%'    FAIL
' 42 '     OK 42         'West'   FAIL
'1e3'      OK 1000       ''       FAIL ...the value  is not a number
'-5'       OK -5
```

That last line is what produced **75e40cf**: refusing `""` contradicted the rule that blank contributes nothing, and `""` is what an export writes for a missing number. It also shows why values are now quoted — `the value  is not a number` has a hole where the value should be.

PR #79, opened 2026-08-07
pipelines: Delete data, deactivated activities, and an honest WebHook

calvinchengx:
Three of the six missing-activity items from the audit's follow-up, each measured before written and each with tests proven able to fail.

### Deactivated activities never run (`dcb3faf`)

`state: Inactive` was silently dropped on parse, so an activity a user explicitly commented out in the portal **executed anyway, side effects included**. Now: the run report carries the reserved `Inactive` placeholder status, branching follows `onInactiveMarkAs` (a `Failed` mark steers `UponFailure` branches **without failing the run** — a placeholder cannot produce an error nobody hit), and no output fields are invented, so a downstream reference fails to resolve exactly as Fabric documents. Mutation-checked: with the gate disabled both tests fail on "deactivated activity executed".

### Delete data activity (`464f7a1`)

Really removes data through the storage layer — so `FileDeleted` events fire and a **Reflex trigger sees a pipeline Delete** exactly as it sees an ADLS client's delete. Implicit directories handled (seeding `Files/d/a.txt` creates no `Files/d` row — the first version's existence probe missed this and its own test caught it); `recursive` removes the subtree, non-recursive removes direct-child files only; missing path **fails loudly naming the path** — held to the loud side while the real oracle is unmeasured, and the code comment says so. The `Delete` type name is ADF-inherited and schema-verifiable, unlike the two activities this PR deliberately does NOT add (see below).

### WebHook stops lying (`e0c58d8`)

`WebHook` silently aliased `Web`, so a webhook pipeline "worked" locally while its **defining half — call back and park — never executed**. Now refused by name, pointing at the supported alternative and at docs/37 §4 (inline pipeline execution cannot park; async pipelines are the prerequisite for the real thing). Parity row split: 🟢 Web / 🔴 WebHook refused.

### Deliberately absent: Refresh SQL Endpoint, Lakehouse maintenance, KQL, SJD wrappers

Their docs pages show **no JSON payloads** — the activity `type` strings are unverifiable, and inventing wire names is the failure this repo's history keeps punishing. Blocked on a one-time capture from a real tenant (portal JSON view of a pipeline containing the four); everything else is ready to wire.

### Verification

- `go test ./internal/api/ ./internal/pipeline/` green; gofmt/vet clean
- `check_witnesses.py --strict`: **86 claims** (Delete's new entry), 0 TODO — including the checker catching that a renamed claim needs its manifest key renamed (`web-activity-and-webhook-refused`), which is the manifest discipline working
- `make check` green
- No changes to `internal/api/jobs.go` — the CopyJob session's territory (`feat/copyjob-executes`); dispatch-switch adjacency in `pipelines.go` flagged to them for merge ordering


calvinchengx:
Scope grew by one commit: **AzureFunctionActivity** (Phase 3 of docs/43), landed here rather than a third PR contending for the same dispatch switch. ADF's published schema is the oracle (method enum, body-on-GET forbidden, body-on-POST/PUT required — each refused by name); the mechanics ride a new shared `httpActivity` core extracted from Web so the two cannot drift; the witness's stand-in function app rejects a wrong `x-functions-key`, so the happy path cannot be satisfied by a server that lets anyone in. Claim count 86 → 87, strict checker green, full package green.

PR #84, opened 2026-08-08
spark jobs: the emulator runs the job, as Fabric's pool does

calvinchengx:
Closes the last genuinely-movable Data Engineering yellow: **Spark Job Definitions** goes 🟢.

### The gap, which I had mis-scoped

I first said this yellow needed "witness depth, not new code" — read off the status cell, not the code. `internal/api/spark_jobs.go` had four functions (parse, save, get, report) and **no dispatch**. `jobs.go` correctly parks every SJD that parses at `CompleteAt=MaxInt64` so the clock can never finish it, and then nothing ran it: the emulator served the resolved job at `GET …/sparkJobRun` and waited for a `POST …/sparkJobRunResult`.

**The only thing that ever executed a Spark job was the e2e**, which fetched the source and `exec()`d it in its own process. Real engine, real computation — but the suite was the engine, not the witness. Real Fabric runs a submitted SJD on its own pool.

### The fix

`internal/api/sparkjobdrive.go`, deliberately the same shape as `driveNotebookRun`:

- **Dispatch** from `jobs.go` beside the notebook branch, gated on `runsNotebooksItself()`.
- **Same silence guard** — a drive that dies without finalising fails the job.
- **Lakehouse + Environment** bound exactly as a notebook run binds them.
- **`sys.argv` = main file + arguments**, set in-session before the source runs.
- **JARs refused before any user code runs** on an engine without a JVM, **asked of the engine** rather than assumed from config.
- **One finalisation path** shared with the HTTP callback.

**With no agent configured, nothing changes** — the callback contract stands, pinned by a test.

### Witnesses

- 5 Go tests: executed-by-emulator (asserting the source *and* the argv prelude reached the agent), terminal-event-on-the-bus, failure-fails-the-job, unreachable-agent-fails-rather-than-parks, no-agent-waits-for-callback.
- **Terminal event pinned with subscribe-then-drain**, not `Replay()` — the drive is async.
- **Mutation-checked**: removing the dispatch fails 4 of 5; the no-agent test correctly still passes.
- `e2e/notebook-run` **stops playing the engine** — it polls the job the emulator drives. `sjd-result=5` is the unfakeable half: the engine read the bound lakehouse (3 rows) and got the definition's arguments (2).

### Also fixed: the row was overclaiming before this change

Its description ended "and real Sail/JVM execution witness" while its status said "parse layer only". And `witnesses.json` credited only a parse test while `ci:notebook-run` exercised the whole loop.

### Verified

`go build`, `go vet`, full `./internal/...`, **`-race -count=2`**, `check_witnesses.py --strict` (89 claims, 0 TODO), `make check`.

### Note on the branch name

Originally branched as `feat/spark-job-execution` — which collided with another session's already-pushed branch of that name (the HDInsight work, since merged as #83). Renamed rather than force-pushed.


calvinchengx:
Closing: this PR's head branch carries the wrong commit entirely.

`feat/emulator-executes-spark-jobs` points at 3740018 — the HDInsight Spark commit, already merged as #83 — not the Spark Job Definition execution work this title and body describe. Cause: I branched and committed in the shared checkout root while another session's branch of a colliding name was checked out there, so my `git branch -m` renamed *their* branch and my own commit landed on a third branch entirely.

The actual work is intact at 8201884 and pushed to `rescue/sjd-execution`. Reopening from there as a new PR; nothing is lost.

PR #85, opened 2026-08-08
pipelines: Databricks, Azure Batch and Azure ML activities

calvinchengx:
Phase 5 continues. Same bet as #83's HDInsight row and the Livy row before it: **terminate the submission contract locally, let the engine the emulator already runs execute the code.**

### Three activities, three ADF discriminators

`DatabricksNotebook` (notebookPath), `DatabricksSparkPython` (pythonFile), `DatabricksSparkJar` (mainClassName) — oracle is the published schema, required fields and all.

### The delivery difference is the point

A **notebook** task's `baseParameters` arrive **bound as names**, the way `dbutils.widgets` delivers them — notebook code reads `run_date`, and a value that only lands in argv breaks it. A **python** task's `parameters` arrive as **`sys.argv`** with argv[0] the file. Opposite mechanisms, and each test asserts the other's is absent — mutation-checked: routing notebook parameters through argv fails `TestDatabricksNotebookRunsWithItsBaseParameters` on "baseParameters are not bound as names".

### Refused by name, each with its cause

- **JAR task** — a Java/Scala main class needs a JVM classpath the default engine (Sail: Rust, embedded CPython) does not have. Same cause as `spark.jars` and HDInsight's `className`. Running a Python file in its place would be inventing behaviour.
- **`libraries`** — installs on a cluster whose lifecycle the emulator does not own; an Environment item is the modelled way (docs/37 §1). Accepting the field and installing nothing is the silent version of the same gap.
- **`dbfs:` / `/Workspace` / `/Shared` / `/Repos` paths** — the emulator has no DBFS or workspace tree. Reinterpreting one as a lakehouse path would invent a mapping nobody wrote; files address OneLake as `<lakehouseItemId>/<path>`.
- No engine attached → honest failure naming it. Output carries `executedBy` rather than a fabricated `runPageUrl` that would 404.

### Verification

23/23 packages, strict checker **90 claims**, `make check` green, gofmt/vet clean.

Branched off `origin/main` deliberately: another session has uncommitted SJD-execution work in the shared root (`internal/api/sparkjobdrive.go` and friends), which my first `checkout -b` aborted on. Untouched, and flagged to them — including that #83's HDInsight path uses the same agent `/statements` seam with a per-activity session name, in case their driver assumes one session per process.


calvinchengx:
Scope grew to the rest of Phase 5's tier plus a coverage pass. Three commits added since review:

### Azure Batch (`Custom`) — **off by default, and that is the design**

Every other compute activity here runs *Python through the Spark agent*: user code in the engine's sandbox. A `Custom` activity's `command` is a **shell command** — a process on whatever host runs the agent. `internal/config/config.go` already states the repo's position on that class, about the terminal pane: *"a terminal is not another read; it is arbitrary execution"*, and *"Empty = the feature does not exist"*.

So this takes the same posture: without `FABRIC_CUSTOM_ACTIVITY=shell` the activity **refuses by name and no command reaches the agent**. That is asserted, not assumed — `TestCustomActivityIsOffByDefault` fails if a single statement is sent with the gate shut, and the mutation that opens the gate prints the `echo pwned` that would have run.

Enabled: the command runs in the **agent's container**, not the emulator's process; `extendedProperties` become environment variables as Batch documents; the command's **own exit code decides the activity**; and an unreadable report **fails**, because an unknown exit status is not success. `resourceLinkedService`/`folderPath`/`autoUserSpecification`/`referenceObjects` are refused by name even when enabled — each would leave a command running in conditions the definition does not describe.

**100% statement coverage on `customActivity` and `lastJSONLine`, zero uncovered lines.**

### The input surface of the other four activities

Coverage measured, not assumed: `functionsActivity` was **71.7%**, `hdinsightSparkActivity` **75.0%**, `webhookActivity` **80.6%**, `databricksActivity` **87.5%**. The gap was the entire input layer — every expression that fails to resolve, every typeProperty of the wrong JSON shape, every unreachable agent. Now **92.5 / 94.1 / 97.2 / 98.6%**, with `parkUntil`, `asInt` and `splitRootPath` at 100%.

Two things that pass found in the writing: parameters resolve *after* the file lookup (half the cases were asserting a not-found error instead of the field under test — caught because assertions name specific fields), and `sparkConfig` and headers had **no test proving they reach the engine or the wire** — a success-only assertion would never have noticed them being dropped.

### Also

The JAR/main-class boundary correction: both refusals said "use the JVM overlay", which does not remedy — nothing submits a main class there either (verified independently: no `spark-submit` anywhere, `/statements` routes non-SQL to `exec(compile(...))`). A JAR *library* attaching is a different capability from *executing* a named main class, and the rows now say so separately.

23/23 packages, strict checker 90 claims, `make check` green.

PR #100, opened 2026-08-08
docs check: derive the published set instead of copying it

calvinchengx:
Found by running the sweep fork 2 proposed — grep the tree for comments telling a future reader to keep two things in step, on the theory that **such a comment is a defect already filed, with no owner and no failing test.** 54 hits, most of them "mirrors" used descriptively. This one was real.

## The coupling

`scripts/check_docs_sidebar.py` decides which docs must appear in the sidebar. Which docs are *published* is decided elsewhere — by `DOC_RE` in `website/scripts/sync-docs.mjs`. The checker kept its own copy of that regex under:

```python
# Mirrors DOC_RE in website/scripts/sync-docs.mjs — the set the site publishes.
PUBLISHED = re.compile(r"^(\d{2}-.*|parity|engine-matrix)\.md$")
```

## Why the existing test did not pin it

There *was* a test named `test_the_pattern_matches_the_sites_own_regex`, whose comment said exactly the right thing — *"If that regex is edited and this one is not, the check silently guards a different set."* It then asserted:

```python
assert "DOC_RE" in mjs, "sync-docs.mjs no longer defines DOC_RE"
```

…that the **name** still appears in the file. Followed by checks of the local copy against a hand-written list of filenames.

Both halves pass for the wrong reason. Edit `DOC_RE` to publish a different set and the name is still present, the hand-written list still matches the stale copy, and the checker silently guards a set the site no longer uses. **The assertion matched a phrase that co-occurs with the claim rather than the claim itself** — the same shape as a refusal test that matched `"JVM overlay"` while the string survived a rewrite as a contrast clause.

## The fix: derive, don't check

Per the selection rule that came out of today's three instances — **derive** when one list is a function of the other, **collapse** when two lists are one question asked twice, **check** only when neither is possible — this is a derive case. The Python pattern was a *copy* of the JS one, so it can be read from it and the second list stops existing.

`published_pattern()` extracts `const DOC_RE = /…/;` and compiles it. It **raises rather than falling back**: an absent file, a missing declaration, or a regex outside the JS/Python shared subset all fail loudly and name the reason. A silent fallback to a default would rebuild precisely the bug being removed.

## Verification

- `make check` — unchanged output, **44 published docs**, same set as before.
- Full `python/tests` — **469 passed, 6 skipped**. `ruff` clean.
- **Mutation-checked with a passing control.** Reverting `published_pattern` to a hardcoded copy fails three tests, the first with `"the site's own regex was not used"`. Control green.
- The new tests pin the derivation, not a restatement: one writes a temp `.mjs` with a *different* `DOC_RE` and asserts the published set changes — an assertion a copied pattern cannot satisfy.

One aside worth recording, because it is the same lesson twice in one change: my first "Python rejects this regex" probe used `(?<=a)` lookbehind, which **both** languages accept, so the test did not raise. The probe is now a JS named group `(?<kind>…)`, which Python genuinely rejects. Asserting the substance applies to the test's own fixtures too.


calvinchengx:
Consolidated into #115 — commit 7d110e3 cherry-picked verbatim, so this lands with one build instead of queueing its own.

It was also blocked here regardless: this branch touches no Go, but inherited main's sub-90% coverage floor failure. #115 fixes that too.

Branch kept, not deleted.

PR #114, opened 2026-08-08
tests: cover the value coercions and the pagination stop conditions

calvinchengx:
Raises real Go coverage, because the 90% floor has **1.2 points of headroom by its own admission** (`ci.yml:192`) and both gated legs just measured **89.7–89.8%** on Go code that measured **90.8%** an hour earlier. #100 tripped it while changing two Python files. A gate with less headroom than its own measurement noise fails on whoever pushes next; re-running until it lands is gaming it.

Local repo total **88.4% → 88.7%** (no SQL Server here, so the warehouse tests skip — CI's gated legs read ~1.2 points higher).

## `internal/semanticmodel`: 92.7% → 94.9%

The value coercions — `truthy`, `dstr`, `asNumber`, `toF` — were at 33–62%, and the unreached arms are not exotic:

- **Booleans** reach them through comparisons. `(1 = 1) & ""` must render `"TRUE"`, and `(1 = 1) + 1` must be `2` — DAX renders booleans TRUE/FALSE and counts them 1/0.
- **`IF`'s condition** was only ever tested with a boolean. Blank, numeric and text conditions all have rules (`FALSE`, `BLANK`, `0`, `""` are false) and none was exercised.
- **Sized integers** reach `toF` through Direct Lake, where rows arrive from parquet as `int32`/`uint8`/`float32` rather than JSON `float64`. Each of the twelve numeric Go types is now summed with a second row, so a dropped value shows as a wrong total rather than a plausible one.
- A value of **no** numeric type must refuse rather than count as zero — the same rule the text refusal enforces.

Driven through real DAX rather than by calling the helpers directly: a coercion that is correct in isolation and unreachable from a query proves nothing about the evaluator.

## `internal/api/restpagination.go`: three selectors 0%/50%/33% → 100%

These decide whether a REST connector takes another page or stops, and the **stop conditions were the uncovered half** — which is the dangerous half, since a selector that wrongly reports "keep going" pages forever against a real API.

- `headerSelector` — every spelling Fabric accepts (`headers.X`, `headers['X']`, `headers["X"]`), plus the malformed ones that must not resolve to the empty header name, which `hdr.Get("")` answers with `""` and a caller could read as a legitimate empty value.
- `resolveValue` — each documented stop: missing JSONPath, explicit `null`, absent header, header present but empty, malformed selector, empty literal.
- `isEmptyNode` — emptiness in each JSON shape, **and** that `0` and `false` are values rather than emptiness. A laxer check truncates a real result set.

## Mutation-checked

Two mutations, each reverted after:

- accept an empty header name (`len(rest) > 1` → `true`) → **fails**
- treat `0` as an empty collection → **fails**

`go build`, `go vet`, full `go test ./...`, and `make check` pass.

## Not touched

Per `CLAIMS.md`: `internal/purview/*` and `internal/store/purview.go` (parity fork), `internal/auth`/`api/admin*` (tenant-admin gate), `pipelines.*` and the activity files (Activities session). `internal/warehouse` and `internal/tds/sqlserver.go` are DSN-gated and only measurable on a leg with an engine.

### DIFF
--- a/internal/api/restpagination_selectors_test.go
+++ b/internal/api/restpagination_selectors_test.go
@@ -1,0 +1,119 @@
+package api
+
+import (
+	"net/http"
+	"testing"
+)
+
+// The pagination selectors decide whether a REST connector takes another page
+// or stops. They were the least-covered code in the connector — `headerSelector`
+// at 0%, `resolveValue` at 50%, `isEmptyNode` at 33% — and every uncovered arm
+// is either a selector spelling Fabric accepts or one of its documented STOP
+// conditions. An unreached stop condition is the dangerous half: a selector that
+// wrongly reports "keep going" pages forever against a real API.
+
+// headerSelector accepts the spellings Fabric's expressions use. The false
+// cases matter more than the true ones: each is a malformed selector that must
+// not resolve to the empty header name, which `hdr.Get("")` would answer with
+// "" and a caller could read as a legitimate empty value.
+func TestHeaderSelectorSpellings(t *testing.T) {
+	for _, tc := range []struct {
+		expr string
+		want string
+		ok   bool
+	}{
+		{"headers.Link", "Link", true},
+		{"headers['X-Next-Page']", "X-Next-Page", true},
+		{`headers["X-Next-Page"]`, "X-Next-Page", true},
+		// Malformed: a name is required, and an empty one is not a name.
+		{"headers.", "", false},
+		{"headers['']", "", false},
+		{`headers[""]`, "", false},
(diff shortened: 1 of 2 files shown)


calvinchengx:
Consolidated into #115, which now carries this branch's `dax_coercion_test.go` verbatim.

Its pagination tests are dropped as redundant: #115 already takes `headerSelector`, `resolveValue` and `isEmptyNode` to 100%, and both branches independently named a test `TestHeaderSelectorSpellings` — whichever merged second would have broken `main` with a redeclaration error that neither PR's CI could see.

Branch kept, not deleted.

PR #116, opened 2026-08-08
test(rest): cover the pagination value layer

calvinchengx:
Follow-on to #115, independent of it — different files, either can merge first.

`restpagination.go` was the worst-covered substantial file in `internal/api` at **74.6%**, and the gap was not spread evenly. `headerSelector` was at **0%**, meaning the header half of what the file's own doc comment promises — *"a JSONPath into the body, or `Headers.x`"* — had never been executed. Cursors arriving in a response header are a documented Fabric shape and were entirely unverified.

The existing `restpagination_test.go` drives whole paging sequences through an httptest server and asserts the request log, which covers the loop well and the value layer barely. This adds direct unit tests for the pure helpers plus three integration tests for the header paths:

- `headerSelector` — all three spellings, and the malformed ones refused rather than read as a header named `""`
- `resolveValue` — JSONPath, null (Fabric's documented stop), header, absent header, literal
- `scalarString` — `1000` not `1000.0`, since every offset arrives as a float64; objects and arrays refused rather than rendered into a URL
- `isEmptyNode` — including that **zero is a value, not an absence**, so an offset of 0 keeps paging
- `splitRuleKey` / `parseRange` — the refusals, including a zero step that would never terminate
- `resolveRelative` — absolute, root-relative, path-relative, query-only, and malformed
- end to end: a cursor read from a response header, a cursor carried into the next request's header, and an `EndCondition` on a header

The assertions are mostly refusals on purpose. A resolver that returned `("", true)` where it should return `false` pages forever against a live endpoint, and no happy-path sequence would reveal it.

**Result:** `restpagination.go` 74.6% → ~99% (every function ≥88.9%); repo total 89.3% → 89.6% locally, so roughly 90.9% on CI once #115 is in.

### DIFF
--- a/internal/api/restpagination_values_test.go
+++ b/internal/api/restpagination_values_test.go
@@ -1,0 +1,345 @@
+package api
+
+// Where a pagination VALUE comes from, as opposed to how the loop steps.
+//
+// restpagination_test.go drives whole paging sequences and asserts the requests
+// the server saw. That covers the loop well and the value layer barely: a
+// cursor read out of a response HEADER — half of what the file's own doc
+// comment promises ("a JSONPath into the body, or `Headers.x`") — had no test
+// at all, and neither did the scalar rendering every rule depends on.
+//
+// These are direct unit tests because the functions are pure, and because the
+// interesting cases are refusals: a value that cannot be resolved must report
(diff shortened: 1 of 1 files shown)


calvinchengx:
Consolidated into #115 — same two test files, one build instead of two.

PR #119, opened 2026-08-09
fix(e2e): seed the tenant these suites expect

calvinchengx:
**`main`'s containerized CI is broken right now.** This unblocks it.

### What happened
`entra-emulator` **v0.4.0** (published Aug 8, 15:35 UTC) moved its default seeded tenant:

```
11111111-1111-1111-1111-111111111111  ->  6f89cf12-978b-4d23-ac18-9ef0c127cf87
```

All 23 containerized suites here pull `ghcr.io/calvinchengx/entra-emulator:latest` **unpinned**, while 62 files still name the old tenant in their issuer URLs. So entra began seeding a tenant nothing referenced, discovery started 404ing, and the token exchange surfaced `HTTP 400` — **29 jobs red**.

The timeline pins it, and it is not any one PR's fault:

| When | What |
|---|---|
| Aug 8, 10:17 | main CI green (e96c2a7) — 22 containerized jobs, 0 fail |
| Aug 8, 15:35 | **entra v0.4.0 published to `:latest`** |
| Aug 9, 00:44 | next PR run — 29 containerized jobs fail |

### The fix
Seed `TENANT_ID` on the entra service in each compose, pinning the tenant that suite already expects. `:latest` stays free to move. Proven against the published image:

```
without TENANT_ID:  GET /11111111-.../v2.0/.well-known/openid-configuration  ->  404
with    TENANT_ID:  same URL                                                 ->  200
```

Same pattern `arm-emulator` already adopted (`fix(compose): seed the tenant arm-emulator expects`).

### Scope
23 compose files, 3 lines each. Verified every file still parses as YAML and that `TENANT_ID` landed on the **entra** service in all 23 — not on a neighbouring service.

Migrating all 62 files to the new GUID is the bigger, longer-term alternative; it can happen on its own schedule now that CI is unblocked. #118 (the yamlEscape fix) is blocked behind this and should go green once this lands.

### DIFF
--- a/e2e/airflow/docker-compose.yml
+++ b/e2e/airflow/docker-compose.yml
@@ -15,6 +15,9 @@ services:
   entra-emulator:
     image: ghcr.io/calvinchengx/entra-emulator:latest
     environment:
+      # entra >=v0.4.0 seeds a different tenant by default; pin the one
+      # this suite's issuer URLs name, or discovery 404s.
+      TENANT_ID: "11111111-1111-1111-1111-111111111111"
       ORIGIN_MODE: compat
       TLS_ENABLED: "false"
       PORT: "8443"
--- a/e2e/azurite-shortcut/docker-compose.yml
+++ b/e2e/azurite-shortcut/docker-compose.yml
@@ -14,6 +14,9 @@ services:
   entra-emulator:
     image: ghcr.io/calvinchengx/entra-emulator:latest
     environment:
+      # entra >=v0.4.0 seeds a different tenant by default; pin the one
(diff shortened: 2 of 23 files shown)


calvinchengx:
Superseded by #120. Seeding `TENANT_ID` fixed only the first layer — it moved the failure from HTTP 400 to 401, because entra v0.4.0 also re-seeded the daemon client id this suite names 67 times, and there is no env knob for app ids. entra's release notes call for migrating the seeded object ids too, and warn against mixing old and new. #120 does that migration and is verified locally against the published image.

PR #120, opened 2026-08-09
fix(e2e): migrate to entra v0.4.0's seeded GUIDs

calvinchengx:
**Unblocks `main`** — its containerized CI is currently red. Supersedes #119.

### What broke
`entra-emulator` **v0.4.0** re-seeded *every* placeholder GUID with a realistic fixed v4 GUID — a documented breaking change. All 23 containerized suites here pull `entra-emulator:latest` **unpinned**, so when v0.4.0 published (Aug 8, 15:35 UTC) 29 jobs went red.

It broke in two layers, which is why the first attempt (#119, seeding `TENANT_ID`) wasn't enough — it moved the error from **400 → 401** but no further:

1. Discovery **404**'d on the old tenant → surfaced as `HTTP 400`
2. With the tenant reachable, the old **daemon client id no longer existed** → `HTTP 401`

entra's release notes call for exactly this migration: *"update every issuer/authority URL embedding the tenant, **and every seeded object ID your tests name**"*, and warn *"do not mix"* — which is what #119 was doing.

### The sweep

| | old | new | count |
|---|---|---|---|
| tenant | `11111111-…` | `6f89cf12-978b-4d23-ac18-9ef0c127cf87` | 105 |
| daemon app | `cccccccc-…-0002` | `00d88624-f0d7-46f6-a641-6232c2608928` | 67 |
| user Alice | `aaaaaaaa-…-0001` | `df8ec5dd-1599-45ef-908b-4ae020cd1dbe` | 1 |

### The part that needed care
`11111111-1111-1111-1111-111111111111` does **double duty** here — it is entra's tenant *and* an arbitrary fixture id for our own objects, where it means nothing to entra. A blanket find-replace would have silently rewritten unit-test fixtures. Six occurrences are deliberately left as-is:

- `internal/api/adxcommandactivity_test.go` — an `itemId`
- `internal/api/mlv_test.go` — a view `iid`
- `internal/api/tenantadmin_test.go` ×3 — domain ids
- `python/tests/test_notebookutils_shim.py` — a workspace id
- `docs/32-xmla-plan.md` — an XMLA `serverName` sample whose meaning I could not establish, so I left it rather than guess

Confirmed after the sweep that no *other* `11111111-*` GUID family was touched, and `go build`/`go vet` still pass.

### Verified locally, not just in CI
`e2e/environment` — one of the 29 failures — run against the **published** `entra-emulator:latest`:

```
client exited (0)
OK: without an Environment, `import cowsay` fails as it must
OK: `import cowsay` succeeds only with the Environment bound — the item reached the session and installed it
```

### Follow-up worth considering
The root fragility remains: 23 composes track `entra-emulator:latest` unpinned, so any future entra release can break CI the same way. [azure-emulators](https://github.com/calvinchengx/azure-emulators) holds the family's certified BOM — pinning against it would make these breaks a deliberate upgrade rather than a surprise.


calvinchengx:
Pushed a second commit: **`go.mod` → entra v0.4.0**.

The first CI run showed the GUID migration fixed only half the suites (containerized went 41 → 44 pass) while the **host-run** ones started failing with `HTTP 400`. Cause: fabric reaches entra through two channels that had drifted apart.

| Suite type | Gets entra from | Seeds |
|---|---|---|
| containerized (23) | `entra-emulator:latest` → v0.4.0 | **new** |
| host-run | `go install` at the version in `go.mod` → v0.3.0 | **old** |

So migrating the fixtures stranded the host-run half against an entra that has never heard of `6f89cf12-…` — the "do not mix" case in reverse.

`e2e/entra_install.py` deliberately reads the version from `go.mod` (there is a long comment on why call sites should not each carry `@v0.3.0`), so this one pin moves every host-run suite onto the same seeds the containers already use.

Go tests pass on v0.4.0 — **24 packages, 0 failures**. They read the tenant from entra's exported constants (`emu.TenantID`, `entra.TenantID`) rather than hardcoding it, so they follow the bump automatically. That also confirms the sweep was right to leave `internal/`'s five `11111111-…` alone: those are fabric's own fixture ids, not entra's tenant.

PR #135, opened 2026-08-09
security: bound range(), request the validated vault URL, stop logging a secret

calvinchengx:
Four of the five open code-scanning alerts. The fifth is a false positive and is dismissed separately, with its reason recorded on the alert.

## #62 `go/uncontrolled-allocation-size` — the one with teeth

`internal/pipeline/funcs.go`, `range(start, count)`: `count` comes from a pipeline definition and `make([]value, 0, count)` reserves the capacity immediately.

I checked what actually happens at both extremes, and the alert names the less dangerous half:

- **Negative** count panics `makeslice`. `evalExpr` already recovers panics into errors by design, so this was merely rude — never a crash. I nearly reported it as one.
- **Huge** count does **not** panic. It allocates: 2e9 × `any` is ~32 GB, and an **OOM kill is a fatal that no per-request recover can catch**. One malformed definition takes the process down with every other in-flight request.

That is precisely the shape `internal/semanticmodel`'s `maxMeasureDepth` exists for, and its comment says the same thing about stack overflow. So `range` gets the same treatment: `maxRangeCount = 100_000`, far past any real pipeline (range feeds ForEach, which Fabric caps well below), plus the `arity` check the case was missing — `args[1]` was indexed unguarded.

Regression test asserts the **error**, not the timing: a test that really tried 2e9 would pass either by OOMing the runner or by having enough RAM, and neither outcome is about the guard. Mutation-checked — removing the bound fails it.

## #60 `go/request-forgery` (critical) — validate one thing, request another

`internal/akv/akv.go`. `checkVaultURI` is a careful allowlist: no userinfo, https enforced on Azure domains, suffix match anchored with a non-empty-label check. It returns the parsed `*url.URL` — and the caller **discarded it** (`if _, err := ...`), then rebuilt the request URL from the raw `vaultURI` string.

The alert is right, and not only about taint tracking. Validating one representation and requesting another is how an allowlist gets bypassed: anything `url.Parse` normalises away is checked in the parsed form and re-introduced by the raw string. They agree today; nothing enforced that they keep agreeing. Now the request is built from `base.ResolveReference(...)` on the URL the check returned.

## #57 `py/insecure-protocol` — `e2e/xmla/run.py`

`PROTOCOL_TLS_SERVER` negotiates the best version *both* sides accept, which still leaves TLS 1.0/1.1 available if the client offers nothing better. ADOMD.NET does far better, so `ctx.minimum_version = ssl.TLSVersion.TLSv1_2` costs this capture listener nothing and stops the suite modelling a downgrade nobody wants.

## #59 `py/clear-text-logging-sensitive-data` — `e2e/notebookutils/notebook.py`

`print(f"credentials.getSecret: {secret!r}")` echoed a Key Vault value into a public CI log. The **assert is the witness**; the print proved nothing the comparison does not. It is a fixture secret, but this file is what a reader copies when wiring up their own vault, so the pattern is the problem. Now asserts and reports a length.

## Not fixed: #58 `py/incomplete-url-substring-sanitization`

`scripts/check_entra_install.py:144` is `check("the final error lost the diagnosis", error and "sum.golang.org" in error)` — a **test assertion that a diagnostic string survived retry exhaustion**. It sanitises no URL and authorises nothing. Dismissing as a false positive rather than contorting the assertion to satisfy the pattern.

`go build`, `go vet`, `go test ./...`, `make check`, and the Python suite (506 passed) all pass.

### DIFF
--- a/e2e/notebookutils/notebook.py
+++ b/e2e/notebookutils/notebook.py
@@ -47,8 +47,11 @@ assert "greeting.txt" in listing and "copy.txt" in listing, listing
 
 # --- credentials.getSecret: brokered read from Key Vault ---------------------
 secret = credentials.getSecret("db-vault", "db-password")
-print(f"credentials.getSecret: {secret!r}", flush=True)
-assert secret == "s3cr3t-value", secret
+# The ASSERT is the witness, not the print. Echoing the value proved nothing the
+# comparison does not, and it put a secret in a public CI log — a fixture one
+# here, but this file is what a reader copies when they wire up their own vault.
+assert secret == "s3cr3t-value", f"unexpected secret, {len(secret)} chars"
+print(f"credentials.getSecret: matched the expected value ({len(secret)} chars)", flush=True)
 
 # --- notebook.run: schedule a child notebook through the control plane -------
 # `run` returns the child's EXIT VALUE, per Fabric's reference — not the job
--- a/e2e/xmla/run.py
+++ b/e2e/xmla/run.py
@@ -605,6 +605,11 @@ log("issuing a self-signed CA for the container's view of this host")
 make_cert(WORK)
 
 ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
+# PROTOCOL_TLS_SERVER negotiates the best version BOTH sides accept, which still
+# leaves TLS 1.0/1.1 on the table if the client offers nothing better. ADOMD.NET
+# does far better than that, so pinning the floor costs this listener nothing and
+# stops the suite modelling a downgrade nobody wants.
+ctx.minimum_version = ssl.TLSVersion.TLSv1_2
 ctx.load_cert_chain(os.path.join(WORK, "cert.pem"), os.path.join(WORK, "key.pem"))
 srv = TLSServer(("0.0.0.0", PORT), Capture)
 srv.socket = ctx.wrap_socket(srv.socket, server_side=True)
--- a/internal/akv/akv.go
+++ b/internal/akv/akv.go
@@ -100,11 +100,21 @@ func New(insecure bool, client *http.Client, extraHost string) *Client {
 // ResolveSecret GETs {vaultURI}/secrets/{name}?api-version=… with the bearer
 // token and returns the secret value.
 func (c *Client) ResolveSecret(vaultURI, name, bearer string) (string, error) {
-	if _, err := c.checkVaultURI(vaultURI); err != nil {
+	// Build the request from the URL checkVaultURI RETURNED, never from the raw
+	// argument. Validating one representation and then requesting another is how
+	// an allowlist gets bypassed: any input `url.Parse` normalises away (a stray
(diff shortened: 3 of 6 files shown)


calvinchengx:
Pushed 9e72bef, which fixes a regression this PR introduced. The remaining CodeQL failure is a pre-existing alert, not a new one — detail below.

## 1. Path traversal, introduced by the `ResolveSecret` rewrite (fixed)

Moving the URL build to `ResolveReference` dropped the `url.PathEscape(name)` that was there before. The name is caller-supplied, arriving in the AKV-reference connection body next to the `vaultURI` the allowlist constrains, and `/` is legal inside a decoded `url.URL.Path` — so it stays a separator, and `ResolveReference` then removes dot segments on top:

| name | before this PR | on 8ef1ea8 |
|---|---|---|
| `db-password` | `/secrets/db-password` | `/secrets/db-password` |
| `../../certificates/evil` | `/secrets/..%2F..%2Fcertificates%2Fevil` | **`/certificates/evil`** |
| `a/b` | `/secrets/a%2Fb` | `/secrets/a/b` |

Constraining the host is only half the job: the request carries a vault-audience bearer token, and `/certificates` and `/keys` sit on the same host behind that same token. So this widened what a connection body can reach.

Fixed by setting `RawPath` alongside `Path` and building the URL directly instead of through `ResolveReference`, which keeps the PR's actual goal — request the URL the allowlist returned — while restoring single-segment escaping.

`TestASecretNameCannotLeaveTheSecretsPath` covers it and fails on 8ef1ea8:

```
--- FAIL: TestASecretNameCannotLeaveTheSecretsPath
    akv_test.go:194: "../../certificates/evil" escaped the secrets path: /certificates/evil
```

## 2. The CodeQL failure is alert #60, which is already open on main

`go/request-forgery` at `internal/akv/akv.go`. It is [open on `refs/heads/main`](https://github.com/calvinchengx/fabric-emulator/security/code-scanning/60) at line 112 and this PR moves the same code to line 122, so the PR check reports it as "new alert in code changed by this pull request". No new exposure was added.

It also cannot be fixed by restructuring. The taint is `vaultURI` → `checkVaultURI` → the returned `*url.URL` → `http.NewRequest`, and the host genuinely does come from caller input — that is what an AKV reference *is*. The mitigation CodeQL's own help recommends for this rule is an allowlist, which `checkVaultURI` already implements: anchored Azure vault suffixes with a non-empty label, https enforced except for the explicitly configured emulator host, and userinfo rejected. Any rewrite that satisfies the query would have to derive the host from something other than the reference, which is a different feature.

So this looks like a dismissal ("used in tests" is wrong; **false positive** is the honest reason) rather than a code change, in the same spirit as the standing `go/disabled-certificate-check` disposition on the healthcheck subcommand. I have not dismissed it — that is a repo security-state change and it is alert #60 on main, not something this PR should quietly absorb.

Two options, both fine by me: dismiss #60 with that justification and this PR goes green, or leave it and merge with the CodeQL check red on a known pre-existing alert.

PR #164, opened 2026-08-10
docs: what must not be in data-engineering code

calvinchengx:
The standard for one-flag portability: the code is identical across emulator, DEV, QAT and PROD, and every value that differs is resolved outside it. [doc 21](../blob/main/docs/21-real-fabric-toggle.md) says how the resolver works; this says what user code is not allowed to contain.

The framing test is deliberately sharper than "does it work in both places", because string-templating two branches passes that: **can you `diff` the notebook, the pipeline definition and the dbt project between two environments and get nothing back?**

## Three categories, not two

The usual "abstract what varies, hardcode the rest" misses the case that corrupts deployments silently: values that must be **byte-identical** across environments and are neither free nor abstractable.

The canonical one is **`logicalId`** in `.platform`. It is how Fabric links an item in a workspace to its representation in a Git branch, and the docs say it is "essential not to change it in any way." Two workspaces synced from one branch hold the *same* `logicalId` — that is the mechanism, not a collision. Regenerate it per environment and promotion silently creates a duplicate instead of updating.

## Four corrections to the source taxonomy, argued in the doc

1. **`FABRIC_TARGET` is not an environment name.** Target selects the service and credential *kind*; environment selects the workspace, vault and data. They are orthogonal — `real`+DEV is the normal developer case, `emulator`+PROD-shaped config is how you rehearse a promotion. Putting `emulator` in the same enum as DEV/QAT/PROD makes that rehearsal inexpressible.
2. **Display names are environment values when they encode an environment.** The toggle is name-based, which reads as "names are safe to hardcode." Not when the name is `MSF-DEV-DE-*`. Logical name in code, physical name from the resolver.
3. **`abfss://` URIs are derived, not independent.** A OneLake URI is a function of workspace + item GUIDs. Parameterise it separately and you get two sources of truth; the URI is the one that disagrees silently, because it still resolves — to the wrong lakehouse.
4. **Contract and model *versions* are not environment-invariant.** Names are. Versions differ across environments by design during a promotion — DEV validates v3 while PROD serves v2. Filing them as invariant is what turns the first schema change into an outage.

Also documented: **the bootstrap floor.** Exactly two inputs must come from outside — which environment, and a credential obtainable without a secret. This is why a platform must not require `AZURE_CLIENT_SECRET`: a Fabric notebook has none to give, so demanding one means it cannot run in the service it targets. `contoso-data-platform` shipped that defect while restating the toggle contract locally, and its own source records the lesson: *a contract you copy is a contract you get wrong.*

## Compliance snapshot, measured not asserted

| | verdict |
|---|---|
| `contoso-data-platform` | **compliant** — consumes published `fabric-target`, addresses by name, `DefaultAzureCredential` on the real target, publishes `notebook-content.py` via `updateDefinition` |
| `examples/fab-driven` | source format only (`definitions/silver.Notebook/…` + `.platform`), no resolver |
| the four medallion examples | **not compliant** — correct definition envelope, but pinned to the emulator |

Every claim verified against the tree: `0` medallion examples reference `fabric_target`; `examples/contoso-fixtures/common.py` hardcodes the seeded `TENANT`, `CLIENT_ID` and `CLIENT_SECRET` and defaults endpoints to `https://localhost:*`. `examples/README.md` already calls the inversion *"the wrong way round"* — this doc is the standard those examples should be brought to, and does not claim they meet it.

## Scope

Doc plus its sidebar entry. **No code, no examples touched.** The follow-on work — adopting `fabric_target` in the four examples, and a portability gate that fails any example naming a seeded credential or a `localhost` endpoint — is deliberately not in this PR and needs its own notice.

`make check` passes; the sidebar gate sees 47 published docs, all reachable.

### DIFF
--- a/docs/46-environment-abstraction.md
+++ b/docs/46-environment-abstraction.md
@@ -1,0 +1,189 @@
+# 46 — What must not be in data-engineering code
+
+**The rule.** The code is identical across emulator, DEV, QAT and PROD. Every
+value that differs between them is resolved *outside* the code, by one
+resolver, from one input.
+
+The test is not "does it work in both places" — string-templating two branches
+works. The test is: **can you `diff` the notebook, the pipeline definition and
+the dbt project between two environments and get nothing back?** If a promotion
+requires editing an artifact, the artifact carries an environment value and the
+promotion is a rewrite.
+
+This doc says what those values are. [21 — one toggle](21-real-fabric-toggle.md)
+says how the resolver works; this says what user code is not allowed to contain.
+
+## Three categories, not two
+
+The usual framing is "abstract the environment-specific things, hardcode the
+rest." That misses a third case, and the missing case is the one that corrupts
+deployments silently.
+
+| | what it is | rule |
+|---|---|---|
+| **Environment-bound** | differs per environment | resolve outside the code |
+| **Environment-invariant** | the same everywhere *by the team's convention* | may live in code |
+| **Environment-identical** | must be *byte-identical* across environments, and is neither free nor abstractable | preserve exactly; never regenerate |
+
+The third category exists because Fabric has values that are neither yours to
+choose nor safe to vary. The canonical one is **`logicalId`** in an item's
+`.platform` file: it is the cross-workspace identifier linking an item in a
+workspace to its representation in a Git branch, and Fabric's own documentation
+says it is "essential not to change it in any way." Two workspaces synced from
+one branch hold the *same* `logicalId` for the same item — that is the mechanism,
+not a collision. Regenerate it per environment and Fabric stops recognising DEV's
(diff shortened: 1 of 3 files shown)


calvinchengx:
Renumbered **46 → 47** (`docs/47-environment-abstraction.md`), sidebar entry updated with it.

Another session's ledger claim reserves `docs/46-artifact-persistence.md` — different subject, same number. Theirs is not on any branch yet and this PR was already open, so renumbering this side resolves the collision immediately and unilaterally, rather than leaving it to whoever pushes second. 46 is now free for the work that claimed it.

`make check` passes; the sidebar gate sees 47 published docs, all reachable.

PR #172, opened 2026-08-11
fix: end the Sail e2e on its own terms instead of the job's budget

calvinchengx:
**Adopted work, not mine.** This was authored by the session behind ledger entry `local_1b160431`, which has been silent since Aug 10 19:33 while its fix sat uncommitted in the shared checkout. `local_7cab3362` disclaimed it explicitly; `local_87220308` and I agreed I would adopt it, since I hold the freshest measurement. I applied the 8 files unchanged onto current main and verified them — I did not rewrite the diagnosis, and the analysis below is the author's.

Committed from a worktree off `origin/main`; the shared checkout was never used and still sits at `e40388b` with its files intact.

## Why now

The hang **fired on main** at `796d5df`: 67 success + 1 cancelled, the Sail e2e killed at **25m15s against a 25-minute budget**. I re-ran it and main went green 68/68 with the e2e step genuinely executed, which is exactly the trap the author names below — a cancelled check reads as infrastructure noise, so it gets rerun rather than investigated.

One datum I can add: it reproduced on a **push-event** run, where `ci.yml` sets `cancel-in-progress: false`. So concurrency is excluded as a cause, which supports the client-side diagnosis rather than a scheduling one.

## The cause is in pyspark, not Delta, Sail or the emulator

`SparkSession.stop()` calls `client.close()`, whose first act is `ExecutePlanResponseReattachableIterator.shutdown()` — **process-wide, not per session**. It takes a class-level lock, hands the class-level release pool to `ThreadPoolExecutor.shutdown()` (`wait=True`), and holds the lock until the pool drains. Every new query builds one of those iterators and its `__init__` takes that same lock. So a worker thread's `stop()` blocks the **main** session's next query — precisely where the log falls silent — for as long as outstanding `ReleaseExecute` retries take, which pyspark's own policy lets run "at least 10 minutes".

A **losing** writer is what leaves one outstanding, which is why it is intermittent and only bites when the race actually collides. Measured by driving the shipped pyspark classes with no server at all.

Fix: `session.client.release_session()` instead of `session.stop()`.

## Defence in depth, because the last hang said nothing

- `PYTHONUNBUFFERED=1` in the shared python image. In the run that *passed*, all twenty driver prints carried one timestamp, flushed at exit — so a killed run printed nothing and its silence located nothing.
- A per-step watchdog that names what it was waiting on. Note the author's own correction in the comments: the budget check sits *before* the heartbeat gate, because a `continue` for "nothing to report yet" also skipped the budget, making any budget under one heartbeat unreachable. A watchdog that cannot be made to fire is not evidence.
- Timeouts on `compose up/ps/logs/down`, and on the previously untimed `urlopen` calls in the livy, notebook-driven and spark suites.
- **The budgets sum under the job budget on purpose**: 600 + 60 + 3×60 + 180 = 17 minutes against a 25-minute job. The first draft summed to 28, so the guard could itself be cancelled and prove nothing. Change one number and re-add them.

## Also in docs/20

The Delta side is recorded as measured rather than assumed: the losing writer fails **fast, not eventually** — `effective_max_retries` is 0 for a non-blind-append overwrite, so one attempt then `MaxCommitAttempts`, 50 ms from conflict to both sessions closed. The emulator's half is Azure's own contract: `If-None-Match: *` against an existing commit answers 409 `BlobAlreadyExists`, which object_store does not retry. Terminal on both sides, so an unbounded commit retry was never the explanation.

## Verification

`make check` green, all five touched python files compile, compose config valid. The author reported 26 runs before the fix with no hang reproduced and 20 after, plus `make check` and the spark/livy/notebook-driven suites green locally. **I have not independently reproduced the 26/20 run counts** — CI is the next real test, and this PR is where the fix first gets exercised by the job that was hanging.

### DIFF
--- a/docker/python-runtime/Dockerfile
+++ b/docker/python-runtime/Dockerfile
@@ -9,3 +9,10 @@ COPY python/ ./python/
 RUN uv sync --frozen --no-install-project --group "${UV_GROUP}"
 
 ENV PATH="/app/.venv/bin:${PATH}"
+# Every driver on this image writes its progress to a pipe, where python
+# block-buffers stdout: the e2e/sail driver's twenty prints all carried ONE
+# timestamp, flushed at exit, so a run killed mid-flight printed nothing at all
+# and its 24 minutes of silence said nothing about where it stopped. Progress is
+# only worth printing if a hang can be located by it. Last, so the uv layer
+# above stays cached for all 18 stacks built from this file.
+ENV PYTHONUNBUFFERED=1
--- a/docker/spark-agent/Dockerfile
+++ b/docker/spark-agent/Dockerfile
@@ -26,6 +26,12 @@ COPY python/ ./python/
 RUN uv sync --frozen --no-install-project --group sail-delta
 
 ENV PATH="/app/.venv/bin:${PATH}"
+# Same reason as docker/python-runtime, and set here too because the two images
+# are required to share one preamble (python/tests/test_spark_agent_packaging.py):
+# this agent also writes its progress to a pipe, where python block-buffers
+# stdout, so a run killed mid-flight prints nothing and its silence locates
+# nothing.
+ENV PYTHONUNBUFFERED=1
 
 # Runnable with no command and no mount — the whole repair. Everything the
 # agent still needs is environment (SPARK_REMOTE, the Entra and storage
--- a/docs/20-lakesail-engine.md
+++ b/docs/20-lakesail-engine.md
@@ -113,6 +113,18 @@ touches; every claim below is CI-verified against the emulator:
 - **Concurrent overwrite conflicts are rejected** at the Delta-log commit
   boundary: the executable two-session probe observes one successful writer
   and one transaction failure through OneLake's conditional create contract.
+  The loser fails **fast, not eventually** — worth stating because a hung
+  `sail` job was once read as an unbounded commit retry, and it is not one.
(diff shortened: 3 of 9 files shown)


calvinchengx:
CI caught a real defect in the adopted work, and the fix follows the test's own stated intent rather than weakening it.

`python/tests/test_spark_agent_packaging.py::test_the_runtime_images_share_one_preamble` failed on macOS:

```
ENV drifted between the two runtime images:
  spark-agent   = ['PATH="/app/.venv/bin:${PATH}"']
  python-runtime = ['PATH="/app/.venv/bin:${PATH}"', 'PYTHONUNBUFFERED=1']
```

That test exists precisely to catch this: two Dockerfiles install the same environment, the duplication is deliberate (factoring out a base image would put a release-ordering constraint in the agent's build path), and drift is the stated cost being guarded. `PYTHONUNBUFFERED=1` had been added to `docker/python-runtime` only.

Fixed by setting it on **both**, which is also correct on the merits — the spark agent writes its progress to a pipe for the same reason the drivers do, so it wants unbuffered output just as much. Weakening the assertion to tolerate the difference would have removed the guard and left the agent buffered.

Reproduced locally before and after: 537 passed / 1 failed → **538 passed, 6 skipped**, coverage 92.69% against a 92 floor.

Not a flake and not pre-existing — the same job is green on #171 and on main at `aa12202`. Worth noting the original author's tree would have hit this too; it never met CI.

PR #174, opened 2026-08-11
feat: force the documented async outcomes (getDefinition, createItem, git initialize)

calvinchengx:
Generalises #173 one hour after it landed, because a **second** instance of the same defect turned up the same morning. Two makes it a class, and a class deserves one switch rather than one flag per API.

## The sharper case

`createItem` has two documented outcomes, and the emulator could only ever produce one of them **for the exact item type a real tenant was measured taking the other**:

> **Create Warehouse** — "This API supports long running operations (LRO)". Responses: **201 Created** and **202 Accepted**.
> And: *"This API does not support create a warehouse with definition."*

The emulator went async **only when a definition was supplied**. A Warehouse can never carry one. So:

- real tenant, Warehouse create → **202**
- emulator, Warehouse create → **201, always, unreachably**

A client indexing the body got `None["id"]` against the tenant. That is how `local_87220308` found it, running `examples/medallion-pyspark/provision.py` against the real trial — and the failure landed three calls downstream of the assumption that caused it, after a workspace and lakehouse had already been created.

## What changed

`FABRIC_DEFINITION_LRO` becomes **`FABRIC_FORCE_LRO`**: every API documenting both a synchronous and an asynchronous outcome answers the asynchronous one. Covers `getDefinition` (200 vs 202) and `createItem` (201 vs 202).

**The old name still works** — a compose file or CI leg already setting `FABRIC_DEFINITION_LRO` keeps its async path rather than silently losing it, and that is asserted in `TestForceLROFromEnv`. Renaming rather than adding a second flag because the alternative scales to one knob per endpoint, and the thing being modelled is not endpoint-specific.

Still **off by default**: 200/201 are equally legal and are what most calls see. Flipping the default would be a breaking change dressed up as fidelity. The point is that the other half is reachable at all.

## Tests

- `createItem` under the flag: 202 with all three documented headers, poll to `Succeeded`, `/result` returns the created item, **and the item really exists** at its own URL rather than only inside the operation payload;
- `createItem` default: still 201 with a usable body;
- both `getDefinition` paths, unchanged from #173;
- the env test now covers the new name **and** the old one.

One thing the change surfaced in my own tests, worth stating because it is evidence the flag bites: the getDefinition LRO tests began failing when I flipped it, because their **seed helper's item create** started returning 202 against a `StatusCreated` assertion. Fixed by seeding with the flag off and enabling it afterwards, so each test isolates the surface it is about.

`make check` green; `internal/api`, `internal/server`, `internal/config` suites green; witness re-keyed in `docs/witnesses.json` and `--strict` passes.

## Credit

The Warehouse-create measurement is `local_87220308`'s, from a real tenant. They fixed the client side in #170; this is the server side, so the emulator can actually produce the shape their fix now handles.


calvinchengx:
Pushed a third surface, and **narrowed the claim to match what was actually audited** — the original title said "every documented async outcome", which I could not stand behind once I started sweeping.

**`git/initializeConnection`** documents 200 and 202 and says "This API supports long running operations". The emulator answered 200 unconditionally. It is the interesting one of the three because **its LRO carries a real result body** (`InitializeGitConnectionResponse`), unlike `commitToGit`/`updateFromGit` whose operations have none — so it covers the case where the async answer must hand back the same object the sync answer did. The test asserts exactly that: same `requiredAction`/`remoteCommitHash` through the operation as through the 200.

The result is **recomputed at result time** rather than snapshotted at accept time. That is what "initialize in progress" means, and a frozen snapshot would be read as current.

### Audit scope, stated rather than implied

Three surfaces checked against the reference and covered here:

| surface | documented | emulator was |
|---|---|---|
| `getDefinition` | 200 / 202 | 200 only |
| `createItem` | 201 / 202 | 201 unless a definition was supplied |
| `git/initializeConnection` | 200 / 202 | 200 only |

Confirmed already-async and needing nothing: `updateDefinition`, `commitToGit`, `updateFromGit`, capacity assign/unassign, workspace-identity provision/deprovision, deploy stage content, job instances.

Two documented LROs are **not implemented at all**, so they are gaps rather than wrong shapes and are deliberately out of scope: **Load Table** and **`sqlEndpoints/{id}/refreshMetadata`** (the latter already recorded in docs/46).

**This is not a proof that no fourth exists**, and the docs now say so in those words. The reusable part is the method: for any implemented surface, read its Responses table, and if it lists two outcomes check the emulator can produce both.

`make check` green; api/server/config suites green; witness list updated and `--strict` passes.

PR #185, opened 2026-08-11
fix: a dataset's column list became a path segment

calvinchengx:
Found while reshaping the medallion Copy activities to the dataset model a real
Fabric tenant requires:

```
ErrorCode=InvalidParameter, HybridDeliveryException
The value of the property 'formatProperties' is invalid:
'Value cannot be null. Parameter name: formatProperties'
```

Feeding the emulator that full shape ran, reported `Succeeded` — and wrote to
`Tables/[]/bronze_customers`.

`resolveLoc` flattens the Copy side's nested scopes into one lookup, which puts
`datasetSettings.schema` in reach of the `schema` key. In Fabric's dataset model
that is a **column list**, not a namespace; `fmt.Sprint` rendered the empty array
as `[]` and it became a path segment. `annotations` and `structure` are arrays on
the same object and would corrupt any key they collided with, so the fix is
general: a composite value reads as absent, never as its Go rendering.

The test asserts the **path**, not the status — the first version of it passed on
`Completed` while the bytes were in the wrong place.

Reshaping the four examples' pipeline definitions into this dataset model follows
once #183 lands, to keep those four files in one PR.

### DIFF
--- a/internal/api/pipelines.go
+++ b/internal/api/pipelines.go
@@ -729,6 +729,18 @@ func (e *pipelineExecutor) resolveLoc(side string, raw json.RawMessage, resolve 
 		if err != nil || v == nil {
 			return "", err
 		}
+		// Only a SCALAR addresses anything. Fabric's dataset model reuses names
+		// this lookup wants for things that are not names: `datasetSettings.schema`
+		// is a COLUMN LIST, not a namespace, and `annotations`/`structure` are
+		// arrays too. Flattening the scopes puts them in reach, and fmt.Sprint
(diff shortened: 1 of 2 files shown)


calvinchengx:
Folded into #183. The reshaped example definitions carry `schema: []`, so they depend on this fix to land at `Tables/bronze_customers` rather than `Tables/[]/bronze_customers` — shipping them apart would have made #183 green only on the strength of a second PR merging first. Same commit, same test, now stacked where its consumers are.

PR #186, opened 2026-08-11
fix: the operation state was missing three documented fields

calvinchengx:
Found by turning the sweep rule inward. #184 audits which APIs *answer* 202; this fixes the contract they all poll **through** — including the three surfaces #173/#174/#180 just made async.

## The gap

[Get Operation State](https://learn.microsoft.com/en-us/rest/api/fabric/core/long-running-operations/get-operation-state) documents `OperationState` as:

| field | emulator sent |
|---|---|
| `status` | ✅ |
| `error` | ✅ |
| `createdTimeUtc` | ❌ |
| `lastUpdatedTimeUtc` | ❌ |
| `percentComplete` | ❌ |

And the poll response should carry `Location`, `x-ms-operation-id` and `Retry-After`. The emulator set `Location` only on success and never sent the other two.

So a client following the documented flow — sleeping for `Retry-After`, rendering `percentComplete`, logging when the operation started — got **nothing locally and real values on a tenant**. Same family as everything else this week: local green proving nothing about the tenant.

## Computed, not faked

All three are derivable from the emulator's own clock, so none of them is a plausible-looking constant:

- `percentComplete` is the fraction between `CreatedAt` and `CompleteAt`, from the **same clock the status comes from** — so the two can never disagree. A test asserts exactly that, because "Succeeded at 40%" or "Running at 100%" would render as a stuck or lying progress bar.
- `lastUpdatedTimeUtc` is when the state last **changed** — completion once complete, creation before that. Returning `now` would make every poll look like a fresh update and defeat the field.
- A **failed** operation is 100: the work stopped progressing either way, and `status` is what says which outcome.

`id` is kept even though the documented shape has no such field — removing something clients here may already read would be a gratuitous break, and an extra field is harmless where a missing one is not.

## Location moves, and that is the point

The two documented samples differ, and the difference is load-bearing:

```
running:    Location: …/v1/operations/{id}          Retry-After: 20
succeeded:  Location: …/v1/operations/{id}/result   (no Retry-After)
```

That is how a client following `Location` alone reaches the result without constructing a URL. `Retry-After` appears only while there is still something to wait for — telling a finished client to sleep is wrong. An operation with **no** result (`commitToGit`) gets no Location rather than a dangling one, and that case is asserted too.

## Verification

Three tests: every documented field present and the timestamps parseable as RFC3339; progress agreeing with status across running/succeeded/failed; and the headers following both samples including the resultless case.

Mutation-tested, and one attempt was invalid so I redid it: making `percentComplete` a constant `100` turns it red, and so does stopping `Location` moving to `/result` — the subtle one, since every other assertion still passes. My first attempt at a second mutant deleted the `Retry-After` line, which failed to compile on an unused import; a mutation that does not build proves nothing, so I replaced it with one that does.

`make check`, `go vet`, `gofmt` and all Go suites green.

### DIFF
--- a/internal/api/items.go
+++ b/internal/api/items.go
@@ -3,8 +3,10 @@ package api
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/calvinchengx/fabric-emulator/internal/auth"
 	"github.com/calvinchengx/fabric-emulator/internal/store"
@@ -226,7 +228,18 @@ func (a *API) deleteItem(w http.ResponseWriter, r *http.Request, p *auth.Princip
 type operationBody struct {
 	ID     string `json:"id"`
 	Status string `json:"status"`
-	Error  *struct {
+	// The three fields Fabric's OperationState documents and this emulator did
+	// not send. A client rendering progress, or logging when an operation
+	// started, saw nothing locally and real values on a tenant — a divergence
+	// that only surfaces in production, which is the worst place to find one.
+	//
+	// `id` above is NOT in the documented shape. It is kept because removing a
+	// field clients here may already read would be a gratuitous break, and an
+	// extra field is harmless where a missing one is not.
+	CreatedTimeUtc     string `json:"createdTimeUtc"`
+	LastUpdatedTimeUtc string `json:"lastUpdatedTimeUtc"`
+	PercentComplete    int    `json:"percentComplete"`
+	Error              *struct {
 		ErrorCode string `json:"errorCode"`
 		Message   string `json:"message"`
 	} `json:"error,omitempty"`
@@ -238,16 +251,34 @@ func (a *API) getOperation(w http.ResponseWriter, r *http.Request, p *auth.Princ
 		writeErr(w, http.StatusNotFound, "OperationNotFound", "No such operation.")
 		return
 	}
-	body := operationBody{ID: op.ID, Status: op.StatusAt(a.Store.Now())}
+	now := a.Store.Now()
+	body := operationBody{
+		ID:                 op.ID,
+		Status:             op.StatusAt(now),
+		CreatedTimeUtc:     time.Unix(op.CreatedAt, 0).UTC().Format(time.RFC3339),
+		LastUpdatedTimeUtc: time.Unix(op.LastUpdatedAt(now), 0).UTC().Format(time.RFC3339),
+		PercentComplete:    op.PercentCompleteAt(now),
+	}
 	if body.Status == store.OpFailed {
 		body.Error = &struct {
 			ErrorCode string `json:"errorCode"`
 			Message   string `json:"message"`
 		}{ErrorCode: op.FailWith, Message: "The operation failed."}
 	}
-	if body.Status == store.OpSucceeded && op.ResultRef != "" {
-		loc := "https://" + r.Host + "/v1/operations/" + op.ID + "/result"
-		w.Header().Set("Location", loc)
+	// The polling headers, per the documented samples. `x-ms-operation-id` is on
+	// every poll; `Location` MOVES — it names the state URL while the operation
+	// runs and the RESULT URL once it succeeds, which is how a client following
+	// Location alone reaches the result without constructing a URL. Retry-After
+	// appears only while there is still something to wait for; the completed
+	// sample carries none, and sending one would tell a finished client to sleep.
+	w.Header().Set("x-ms-operation-id", op.ID)
+	base := "https://" + r.Host + "/v1/operations/" + op.ID
+	switch {
+	case body.Status == store.OpSucceeded && op.ResultRef != "":
+		w.Header().Set("Location", base+"/result")
+	case body.Status == store.OpNotStarted || body.Status == store.OpRunning:
+		w.Header().Set("Location", base)
+		w.Header().Set("Retry-After", fmt.Sprintf("%d", a.RetryAfterSeconds))
 	}
 	writeJSON(w, http.StatusOK, body)
 }
--- a/internal/api/items.go
+++ b/internal/api/items.go
@@ -224,6 +224,19 @@ func (a *API) deleteItem(w http.ResponseWriter, r *http.Request, p *auth.Princip
 
 // ---- operations ----
 
+// fabricOperationTime is .NET's round-trip format on a Kind=Unspecified
+// DateTime: 7 fractional digits and NO trailing `Z`.
+//
+// MEASURED against a real tenant on 2026-08-11, and it is NOT what the schema
+// suggests — `string (date-time)` reads as RFC3339, which is what this emulator
+// sent first. A client with a strict parser that accepts `2026-08-11T07:23:41Z`
+// rejects `2026-08-11T07:23:41.3765432`, so it would pass here and fail there.
+//
+// The emulator's clock is second-granular, so the fraction is always seven
+// zeros. That is the right kind of inaccuracy: the SHAPE a client parses is
+// exact and only the precision it carries is coarser.
+const fabricOperationTime = "2006-01-02T15:04:05.0000000"
+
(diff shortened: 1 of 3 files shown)


calvinchengx:
Worth recording what this fix does **not** cover, since it would be easy to assume it does.

`local_0cdd48fd` found that `examples/contoso-fixtures/common.py` carries two LRO resolvers that disagree — verified against `origin/main`:

```
:449  op = r.headers.get("x-ms-operation-id") \        # tolerant, falls back to the Location tail
:458  time.sleep(min(float(got.headers.get("Retry-After", "2")), 20))
:480  op = r.headers["x-ms-operation-id"]              # hard index, no fallback
:481  … time.sleep(1)                                  # flat, ignores Retry-After
```

**This PR removes one trigger and not the defect.** The hard index is currently latent — the header is present on both targets — but `create_item` polls once a second regardless of what the service asked for, and Fabric's own sample returns `Retry-After: 20`. Against a tenant that is ~20x the requested rate on an API family whose reference documents `429 Too Many Requests`.

There is a sharper point here that cuts against this emulator: **it could not have taught anyone to honour `Retry-After`, because until this PR it never sent one on a poll.** An example that ignores a header the emulator does not emit will never be corrected by running against the emulator. That is the two-shapes-on-one-surface category again — the emulator being more forgiving than the thing it emulates — and it is why this PR sends `Retry-After` while an operation is still running rather than treating it as decorative.

`common.py` is held by `local_eb00472c` for a boundary audit; I have reported it to them with the suggested fix (make `create_item` call `post_and_wait`, which deletes the divergence rather than adding a branch) and have not touched the file.

calvinchengx:
**Correcting my own comment above.** `local_eb00472c` caught it and I verified rather than accepting: they are right, and the fact was in code I wrote myself.

I said this emulator "could not have taught anyone to honour `Retry-After`, because until this PR it never sent one on a poll." The clause about polls is true and is what this PR fixes. **The sentence as a whole is false**, and the false part is the part that mattered.

`internal/api/items.go:107` is `if body.Definition == nil && !a.ForceLRO` — so a **definition-bearing create has always been an LRO here**, and `create_item` always sends a definition. `startOperation` sets `Retry-After` on that initiating 202 (`api.go:408`). So every example item create has been taking the async branch, against a response that carried `Retry-After`, and polling once a second anyway.

**The header was there to honour and the example ignored it.** The defect was live locally, on the default path, all along — not latent until a tenant saw it.

The comment at `items.go:100` states this in as many words, and I wrote it in #174. I had the fact in front of me and still reached for the tidier story.

What this PR does is narrower than I claimed and still worth having: it makes the **poll** responses teach the same lesson the initiating 202 already taught, so a client that follows `Location` through the operation gets a backoff at every step rather than only the first. It does not make anything newly true.

`local_eb00472c` had the mirror-image error — first reading was that the async branch never runs locally, so both halves were latent. Neither of us checked the shared premise that `createItem` is synchronous here, which is the third instance today of a conclusion resting on an unverified assumption rather than a wrong measurement.

calvinchengx:
Two defects in this PR, both found by local_0cdd48fd's tenant measurements landing in CLAIMS.md while it was open. Both were mine, and both were the direction this repo keeps getting caught by: **emulator-green, tenant-broken.**

**1. `percentComplete` was interpolated. Real Fabric never emits an intermediate value.**

Measured on a Warehouse create: `null` for the whole time the operation runs, `100` once it succeeds. Nothing in between, ever. This PR was computing the fraction between `CreatedAt` and `CompleteAt` and returning things like 47 — and the comment I wrote defending it said *"It is COMPUTED, not faked"*, which answers the wrong question. The value was honestly derived from this emulator's clock and still wrong, because **the tenant does not publish that quantity at all.**

The client consequence is the sharp part: a progress bar animates smoothly here and sits at `null` against Fabric, and the client's null-handling branch never executes locally even once. Now `*int` — `null` while running, `100` when terminal. The key is always present; only the value varies, because a missing key and a null are different wires.

The `100`-on-success half is measured. **`100` on FAILURE is not** — that is my inference, marked as such in the code and in the test, and it needs a tenant reading rather than an argument to change.

**2. Timestamps were RFC3339. A tenant sends .NET round-trip.**

`2026-08-11T07:23:41.3765432` — 7 fractional digits, **no trailing `Z`**. The schema says `string (date-time)`, which reads as RFC3339, which is what I sent. A client with a strict parser accepts one and rejects the other, so it passes here and fails there.

The emulator's clock is second-granular, so the fraction is always seven zeros. That is the right kind of inaccuracy: the shape a client parses is exact, only the precision is coarser.

**Both mutation-tested, and both mutants verified to compile first** — a mutant that does not build proves nothing.

- Interpolation restored as a compiling mutant: caught at every tick (`t+1s: percentComplete = 10`, `t+2s: = 20`, … through `t+9s: = 90`).
- Format reverted to RFC3339: caught on both fields.

**One of the two assertions was self-referential and I fixed that too.** The timestamp test originally parsed the response with `fabricOperationTime` — the same constant the handler formats with — so it passes for any layout the two share, which is every layout. Under the RFC3339 mutant it went green; only the trailing-`Z` check had teeth. The shape is now asserted as a literal regexp against the bytes a tenant actually sent.

Full suite green. `percentComplete` is now the third fabricated-progress-class defect closed this week, after the 12 activity types and `getDefinition`'s 202.

calvinchengx:
**Both of my last two claims were wrong, corrected against a second round of tenant measurements from local_0cdd48fd. `35fc0e6`.**

**1. `percentComplete` is `null` on Failed, not 100. That was my inference and it was wrong.**

I argued 100 on the grounds that the work stopped progressing either way and `status` is the field that says which. Measured on a failed SemanticModel create, the tenant sends `null`. The rule is not "null while running, 100 when terminal" — it is **100 on Succeeded, null otherwise.**

The reason the tenant is right is better than my reason for the alternative: a client branching on `percentComplete == 100` would read a failure as a completion. That is the fabricated-success shape wearing a different hat, which is what this PR exists to remove. Now keyed off `StatusAt`, not off `CompleteAt`.

**2. The fractional width VARIES, so my fixed `.0000000` was wrong, and so was the test asserting exactly 7 digits.**

From the same tenant minutes apart:
```
2026-08-11T07:49:13.5612398   7 digits
2026-08-11T08:00:43.654668    6 digits
```
So this is System.Text.Json trimming trailing zeros, **not** .NET's fixed-width round-trip "O". Go's `9`s trim and its `0`s pad, so the layout is now `.9999999`. Verified against all three measured samples rather than assumed.

**This is the failure mode I flagged one comment ago, in a new place.** I had a fixed-width emitter and a `\d{7}` expectation. They agreed with each other and disagreed with the tenant — green together, wrong together. The regexp is now `(\.\d{1,7})?` and there is a test that parses **the tenant's actual captured bytes**, including the 6-digit one. That test is the thing my previous version could not have had, because its only oracle was itself.

Our clock is second-granular, so every zero trims and we emit no fraction at all. That is the honest rendering of the precision we have, and a shape the tenant genuinely produces on a whole second — better than seven fabricated zeros claiming sub-second resolution we do not carry.

**Also added, all measured present on every tenant response including the terminal one:**
- `blobInfoId` and `resultContentType` — null in all three samples, so the keys are evidenced and the semantics are not. Emitting null is what the evidence supports; inventing values is not.
- `isRetriable` on the error object. The operation's error is **narrower** than the documented `ErrorResponse`: `{errorCode, message, isRetriable}`, no `moreDetails`/`relatedResource`/`requestId`.
- `error` is no longer `omitempty`. A tenant sends `"error": null` on every non-failed poll, and a missing key is a different wire from a null.

Both corrections mutation-tested, mutants verified to compile first: restoring 100-on-Failed is caught, and restoring the fixed `.0000000` fails to parse the tenant's 6-digit sample. Full suite green.

Witness: measured 2026-08-11 by local_0cdd48fd against the trial tenant — a Warehouse create that succeeded and a SemanticModel create that failed asynchronously, both cleaned up after.

PR #190, opened 2026-08-11
fix: the connection API accepted a shape no tenant does

calvinchengx:
Stacked on #183 (its base), because it changes the same examples. Measured
against a real Fabric trial and its own
`GET /v1/connections/supportedConnectionTypes` (321 types), 2026-08-11.

The tenant refused `secret.py`'s connection with two errors at once:

```json
{"errorCode":"InvalidParameter","message":"Object reference not set to an instance of an object."},
{"errorCode":"InvalidParameter","message":"The CreationMethod field is required."}
```

Chasing that down, the payload turned out to be wrong in **four** independent
ways, and the emulator accepted every one:

| # | Sent | Tenant |
|---|---|---|
| 1 | `connectionDetails.type: "RestApi"` | not among the 321 types; the pipeline REST connector is `WebForPipeline` |
| 2 | `connectionDetails.path` | `path` is the **read** shape; a create needs `creationMethod` + `parameters[]` |
| 3 | `credentialType: "AzureKeyVaultReference"` | not in the enum — the ten values are Anonymous, Basic, Key, KeyPair, OAuth2, ServicePrincipal, SharedAccessSignature, Windows, WindowsWithoutImpersonation, WorkspaceIdentity |
| 4 | `vaultUri` + `workspaceId` in the credential | a `KeyVaultSecretReference` is `{connectionId, secretName, version}` — it names a **connection to the vault** |

### Why #4 is the one that matters

A `vaultUri` requires nothing to exist. The old shape could look valid while
addressing nothing, and locally it always did. The real form makes the vault a
connection in its own right (`type: AzureKeyVault`, `AzureKeyVault.Actions`,
parameter `accountName`) whose own credentials reach it, so the examples now
create two connections — Fabric's design, not ceremony.

### What the emulator does now

- `connectionDetails` requires `type`, `creationMethod`, `parameters`; sending
  `path` on a create is refused by name.
- `credentialType` is validated against Fabric's enum. `AzureKeyVaultReference`
  is rejected with a message naming its replacement, because a bare "unknown
  credentialType" leaves the reader exactly where the tenant's "invalid input"
  left us.
- `keyReference` / `passwordReference` / `tokenReference` /
  `servicePrincipalSecretReference` resolve through the connection they name. A
  field and its reference twin are alternatives; both is refused.
- **Responses render the documented read shape** `{type, path}`, path composed
  from the parameters. The repo's own schema-conformance gate caught this the
  moment `creationMethod` became required — the emulator had been echoing the
  request back, so a client could have read fields real Fabric never sends.

### Verified

Medallion e2e exit 0, all 11 steps, `key-reference connection … via vault
connection …` and Copy still landing 102,000 rows. `make check` exit 0,
`go test ./internal/...` exit 0, `make lint` clean.

Docs updated in 03, 04, 07, 09, 12, 13, 21, 28 and the example READMEs. The
v0.20.0 release note keeps the old shape — it accurately records what that
release shipped.


calvinchengx:
**The three red shortcut jobs look like your new validation doing its job, not a regression. Diagnosis from reading both sides — the run is still in progress so `--log-failed` returns nothing yet.**

Failing step in each:
- `e2e/s3/run.py` — Real S3 server (SeaweedFS)
- `e2e/azurite-shortcut/run.py` — ADLS Gen2 against Azurite
- `e2e/external-shortcuts/run.py` — Shortcut target kinds

All three create a connection first, and all three send **no `connectionDetails` at all**. From [e2e/external-shortcuts/driver.py:46](e2e/external-shortcuts/driver.py:46):

```python
connection = request("POST", f"{FABRIC}/v1/connections", {
    "displayName": "anonymous-object-store", "connectivityType": "ShareableCloud",
    "credentialDetails": {"credentials": {"credentialType": "Anonymous"}},
}, fabric_token)
```

And `connectioncontract.go:81` is now `return d, "connectionDetails is required."`. `e2e/s3/driver.py` and `e2e/azurite-shortcut/driver.py` carry the same shape.

**Which means the drivers were relying on exactly the leniency this PR removes.** They pass today because the emulator accepts a create that a tenant rejects — so these three suites have never exercised the real contract, and their green was the same kind of green the parity map gives: real signal, wrong question.

So the fix is the three drivers, not the validation. Loosening `connectionDetails is required` back would restore green and restore the defect.

Worth saying because it is the argument FOR this PR rather than against it: the moment you made the emulator honest, three suites that had been passing on a shape Fabric refuses went red on the first run. That is the check finding real client-side debt, which is the direction [docs/parity.md](docs/parity.md) structurally cannot see.

I have not run them — containerized, and the logs are not out yet — so treat this as read-not-executed. Happy to take the three driver updates if you would rather stay on the contract itself; say the word and I will branch off #190.

PR #245, opened 2026-08-15
feat: pin SWITCH against Desktop goldens

calvinchengx:
## Summary
- Phase 3: `SWITCH` in the Go evaluator, replayed by `TestDesktopFunctionGoldens` with empty `FABRIC_DAX_URL`.
- Desktop: first match wins; unpaired trailing arg is else; no match without else is BLANK (SUMMARIZECOLUMNS drops the row).
- Equality is not `daxCmp`: BLANK matches only BLANK, never 0 (`SWITCH(0, BLANK(), 10, 99) = 99`). `1.0` matches `1`.

Independent of #244 (`INT`). Branched from `main` at #242.

## Test plan
- [x] `go test ./internal/semanticmodel/`
- [ ] CI `test` job (includes DAX goldens)
- [x] Live pump at `192.168.64.3:8080` agreed on the pinned probes


calvinchengx:
Superseded by #260, which is now on main.

PR #250, opened 2026-08-15
Pin entra-emulator 0.8.1.

calvinchengx:
## Summary
- Additive entra 0.8.1 pin so the family BOM and this repo's compose defaults agree.
- No seed or tenant change.

## Test plan
- [ ] Existing e2e still green against 0.8.1


calvinchengx:
Re-running CI now that the v0.25.0 release has published.

PR #277, opened 2026-08-16
Drop the localRelationSizeLimit preset the pinned engine no longer needs

calvinchengx:
Every agent and runner preset `spark.sql.session.localRelationSizeLimit` to
64 MiB at session start, because an earlier engine served that config as the
string `"3GB"` and pyspark's client calls `int()` on it.

The pinned engine serves it as a byte count:

```
served: '3221225472'
int() parses: 3221225472
createDataFrame without any preset: [Row(id=1, n='a')]
```

Measured against the image `docker/sail/Dockerfile` actually builds — not a
hand-rolled one. A probe image built without `pyspark-client` installed reports
every such config missing and invents a break that does not exist; I know
because I did exactly that earlier and drew the wrong conclusion from it.

## It was not merely dead code

The preset capped local relations at **64 MiB where the engine offers 3 GiB** —
a 48× reduction applied to user code, for a bug that is fixed.

## Why all six sites, not one

They were copies of each other. `engine-matrix/probes.py` said so outright:

> `python/spark_agent/agent.py` presets this for every Connect session … so the
> probe applies the same preset the real runtime does — otherwise it reports
> gaps the emulator does not have.

Removing only the agent's copy would have made that justification false and
left the probe presetting something the runtime no longer does — reporting the
emulator as *more* capable than it is. The two docs claims that cited the
preset move with them, including the 🟢 parity row.

## One thing that needed care

`e2e/eventstream/job.py`'s copy was an `if remote:` block whose `else:` set the
JVM log level. Deleting the block outright orphaned the `else:` and broke the
file — caught by a syntax check, and preserved here as `if not remote:`.

## Verification

- 912 unit tests pass
- `e2e/sail` witness passes end to end
- `e2e/engine-matrix`, which owns the `create_dataframe_local_rows` probe, was
  still running when this was opened — I will confirm it on this PR rather than
  assume it

`e2e/sail/driver.py` still mentions the old parse failure in a comment; that is
being corrected separately in #275.


calvinchengx:
The `engine-matrix` witness finished — the probe that directly exercises the removed preset passes:

```
spark.create_dataframe_local: pass
```

`createDataFrame` works with nothing preset, which is what this PR claims.

No verdicts changed anywhere else in the matrix. Regenerating it produced exactly one diff, on a row that was already ❌:

```
- MERGE INTO delta.`path`  ❌ attribute ObjectName([Identifier("#0")]) is missing from the schema...
+ MERGE INTO delta.`path`  ❌ Table not found: [TABLE_OR_VIEW_NOT_FOUND] ... /tmp/probe/t_merge_pat
```

Different error text from the engine on the same failing verdict — the committed baseline predates the current engine pin. Unrelated to this change, so the regenerated `docs/engine-matrix.md` is deliberately **not** included here; it would mix an engine-version artifact into a config-removal PR.

PR #278, opened 2026-08-16
Publish the compute images under emulator-* names, tagged by dependency

calvinchengx:
Step 1 of the rename: **publish** the new names. No consumer moves in this PR.

## The problem

The sidecar images are named for the repo that publishes them, which is not what
a consumer pins them *for*. `fabric-emulator-sail` carries upstream Sail, and
`databricks-emulator` consumes it with no Fabric in the picture at all.

The tag compounds it. `0.25.0` is fabric-emulator's release number; the Sail
inside is **0.6.6**. Both `0.22.0` and `0.25.0` contain the same engine and
nothing in either tag says so — a confusion that cost me an entire false finding
earlier today, when I read a tag as an engine version and "discovered" a Sail
regression that did not exist.

## What changes

Both images now also publish as `emulator-sail` and `emulator-spark-agent`,
tagged with the version of the dependency each is pinned for:

```
emulator-sail:0.7.0           # the pysail pin
emulator-spark-agent:4.2.0    # the pyspark-client pin
```

Both read from `pyproject.toml` via `scripts/image_tags.py` rather than typed
into the workflow — a tag written in `release.yml` is a second copy of a number,
and the release that bumps `pysail` without editing the workflow publishes a tag
naming the version it no longer contains. Nothing fails; the tag just lies.

**Both names publish this release.** The old name keeps its release-version tag,
so the 21 references across five repos keep resolving while they move.

## The trade this makes, and the gate that pays for it

A dependency tag is deliberately **not an identity**. Both images also carry
first-party code — `launcher.py`, the whole `spark_agent` package — that changes
independently of those pins, so successive releases republish one tag over new
content.

Under the old scheme the tag was the release version, so a bare tag was
effectively immutable and a missing digest cost nothing. That is no longer true:
an unpinned reference now *floats silently* — the compose still parses, the
stack still starts, and the thing under test changed.

`scripts/check_image_digests.py` makes that a build error. The release version
stays recoverable from the OCI labels the metadata action already emits, so the
tag loses that information but the image does not.

## One bug worth flagging

The checker's regex needed a lookbehind. Without it `fabric-emulator-spark-agent`
matches as a suffix of `emulator-spark-agent`, so the gate reported the **old**
name — the one reference this rule explicitly does not govern. Caught by running
it, not by reading it; the comment in the source says so, so the next person to
touch that pattern knows why the lookbehind is there.

## Verification

- `image_tags.py` reads `0.7.0` / `4.2.0` from the pins
- the gate passes on the current tree, and fails on a planted unpinned
  `emulator-sail:0.7.0` reference — checked in both directions
- 912 unit tests pass; both workflow files parse

## Next

Once this releases and the images exist: move the 21 consumer references to the
new names **with digests**, repo by repo, then retire the old name.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -458,6 +458,7 @@ jobs:
       # right, the cast of services had drifted. Nothing about adding a service
       # reminds anyone to redraw, so the roster is asserted instead.
       - run: uv run --frozen --no-sync python scripts/check_arch_services.py
+      - run: uv run --frozen --no-sync python scripts/check_image_digests.py
       # Starlight's sidebar is an explicit list — no autogenerate. A doc left
       # out still builds and still has a URL, but nothing links to it, so it is
       # published and unreachable. Same drift, same guard.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,11 +125,24 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      # The dependency version this image is tagged with, read from the pins
+      # so a pysail bump cannot publish a tag naming the version it no longer
+      # contains. See scripts/image_tags.py.
+      - name: Dependency version for the tag
+        id: dep
+        run: echo "version=$(python3 scripts/image_tags.py ${{ matrix.name }})" >> "$GITHUB_OUTPUT"
+      # Two names during the rename. The new name says what a consumer pins the
+      # image FOR; the old one keeps every existing reference resolving until
+      # they have moved. The old name keeps its release-version tag, because
+      # that is what those references already pin.
       - uses: docker/metadata-action@v6
         id: meta
         with:
-          images: ghcr.io/${{ github.repository }}-${{ matrix.name }}
+          images: |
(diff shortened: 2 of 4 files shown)


calvinchengx:
Pushed `8447d78` — the digest gate had two faults, both found by running it rather than reading it. Fixing them here rather than merging and following up, since shipping a checker I know is wrong then correcting it is the wrong order.

**By file.** It scanned `.md`, so prose naming the image needed a 64-hex digest to pass:

```
README.md:1  emulator-sail:0.7.0   ->  EXIT=1
```

Unreadable, and stale every release. Prose pulls nothing, so the scope is now compose and env only.

**By line.** Some tags float deliberately — `docker-compose.override.yml` uses `:latest`, contoso's `terminal.yml` uses `:dev`, `e2e/fab-driven` builds locally. None has a digest to pin. Those lines now carry `digest-exempt: <reason>`, with the reason **required** and sitting on the line it excuses rather than in a list that drifts away from it.

Without that second fix the migration is blocked outright: three consumer references can neither be pinned nor edited past the gate.

Checked in five directions — prose ignored, unpinned compose flagged, pinned compose accepted, exempted-with-reason accepted, exemption-without-reason **still flagged** so the marker cannot become a silent bypass. The old name stays unflagged throughout.

912 unit tests still pass.

PR #284, opened 2026-08-16
Use KQL that real Kusto accepts in the ADX command activity tests

calvinchengx:
`TestADXCommandReachesTheRealEngine` used

```
.create table Events (ts:datetime, kind:string)
```

which real Kusto rejects. Verified against Microsoft's kustainer in the e2e/rti suite: HTTP 400, `KustoBadRequestException`, `Syntax error: SYN0002: A recognition error occurred. [line:position=1:43]`. Position 43 is `kind`, a KQL keyword and so not usable as a column name.

The Go test passed anyway because `attachEngine` records the command string without parsing KQL, so the assertion (the command reached the engine, against the isolated engine database) held while the command itself could never work. Not a product defect: the emulator relayed the command and reported the engine's 400 faithfully. It is a test-fixture problem, and these tests read as the reference for what an `AzureDataExplorerCommand` activity should contain, so anyone copying one got a 400.

Changes:

- The `.create table` command now uses `(DeviceId:string, At:datetime)`. Those identifiers are proven against kustainer in `e2e/rti/driver.py:122`.
- `TestADXCommandRefusesAQuery` had the same problem in its query, `Events | summarize count() by kind`; now `by DeviceId`. That query is refused before it reaches the engine, so it was never executed either way, but it is just as much a reference.
- A comment above `adxPipeline` noting that the fake does not parse KQL, so commands here should still be valid KQL.

The remaining commands in the file (`.show tables`, `.drop table Nope`) use no keywords as identifiers and are unchanged.

Test plan:

```
go test ./internal/api/ -run TestADXCommand
```

passes. The assertions are untouched: the prefix match on `.create table Events` and the refusal path's `summarize` check both still apply.

### DIFF
--- a/.nwave/des-config.json
+++ b/.nwave/des-config.json
@@ -1,0 +1,7 @@
+{
+  "update_check": {
+    "last_checked": "2026-08-16T14:50:24.333200+00:00",
+    "skipped_versions": [],
+    "frequency": "daily"
+  }
+}
--- a/internal/api/adxcommandactivity_test.go
+++ b/internal/api/adxcommandactivity_test.go
@@ -14,6 +14,14 @@ func adxPipeline(tp string) string {
       {"name":"Cmd","type":"AzureDataExplorerCommand","typeProperties":{` + tp + `}}]}}`
 }
 
+// attachEngine's fake records the command string without parsing KQL, so an
+// invalid command passes here just as happily as a valid one. Keep the commands
+// below accepted by real Kusto anyway: these tests read as the reference for
+// what an AzureDataExplorerCommand activity should contain, and a command that
+// earns a 400 from the engine is a poor thing to copy. `kind` and its fellow
+// KQL keywords are not usable as column names; the identifiers used here are
+// the ones e2e/rti/driver.py proves against Microsoft's kustainer.
+
 // TestADXCommandReachesTheRealEngine: the command the definition names is the
 // command the engine receives, against the ISOLATED engine database for that
 // KQL Database item — not the display name, and not some other item's. A test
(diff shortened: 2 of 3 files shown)


calvinchengx:
Added the KQL validity check to the fake engine (68f65e4), so the fixture fix above is now enforced rather than documented.

`kqlSchemaKeyword` in `internal/api/kql_test.go` reads the `(name:type, ...)` groups that `.create table`, `.create-merge table` and `datatable` share, and refuses a group whose column name is a KQL keyword. The fake answers 400 in kustainer's shape (`KustoBadRequestException`, `SYN0002`). Checked that it bites where the bug lived: restoring the old command makes `TestADXCommandReachesTheRealEngine` fail with the same complaint CI got from kustainer.

Two deliberate limits, both in comments at the code:

- **The keyword list is conservative and partial.** A keyword listed wrongly makes the fake reject KQL real Kusto accepts, which is a worse fault than the one being closed, so absence from the list is not evidence a name is legal. `kind` is the witnessed member; the rest come from Microsoft's identifier-naming rules and are documented rather than proven. kustainer is amd64-only, so the list could not be verified on an arm64 host; proving it in `e2e/rti/driver.py` is the way to upgrade it.
- **No `[line:position=...]` in the error.** kustainer reported position 43 for `kind` in that command, but the token starts at 36 by a plain scan of the text, so the offsets are computed some other way. A position invented here would be one more thing that reads true and is not.

The gate only reads schema declarations. Query bodies, table names and database names are still left to the real engine.

calvinchengx:
Superseded by #287. The branch was renamed from `claude/trusting-leavitt-702e09` to `fix/adx-command-test-kql` to match the repo convention (`fix/jvm-oracle-import`, `fix/portability-gate-in-ci`). Both commits are carried unchanged; `.nwave/des-config.json` was dropped, since it is gitignored and does not exist on main.

PR #327, opened 2026-08-19
fix(spark-agent): rebuild the Spark session the engine dropped

### LINKED ISSUE #312: spark-agent never re-establishes a dropped Spark Connect session
## Summary

The agent builds its `SparkSession` once, at module import, and holds it for the life of the process. If the engine drops that session, the agent never re-establishes it — every statement from then on fails with

```
pyspark.errors.exceptions.connect.IllegalArgumentException:
invalid argument: session <uuid> is not running
```

and keeps failing until the container is restarted. Restarting `spark-agent` clears it every time.

`python/spark_agent/agent.py`:

```python
_b = SparkSession.builder.appName("livy-agent")
if os.environ.get("SPARK_REMOTE"):
    spark = _b.remote(os.environ["SPARK_REMOTE"]).getOrCreate()   # line 34, at import
```

That is the only `getOrCreate()`, and nothing re-runs it. The module docstring states the design plainly — "The agent holds one long-lived SparkSession" — so this is a gap in the recovery path rather than a surprise in the code.

Worth noting that the neighbouring code already anticipates the engine going away underneath. `connectconf.apply()`'s docstring:

> Per session, not once at import: the conf lives on the SERVER, so if the engine restarts while this agent keeps running, the client reconnects to a fresh engine and any earlier normalisation is gone

So the conf normalisation is careful about a restarted engine; the session itself is not.

## What I observed

Twice in one working session, against `databricks-platform-jobs` (Sail-backed):

- Bronze and silver ran fine over the full dataset (102,000 POS customers, 226,544 web order lines, ~474k gold rows).
- The next step — eight `CREATE TABLE IF NOT EXISTS … USING delta LOCATION …` statements registering silver in UC — failed **all eight**, each with `session … is not running`.
- `docker restart <spark-agent>` and re-running: all eight `SUCCEEDED`, no other change.

The gap between the last successful statement and the first failure was a few minutes of idle time in both cases, which suggests an idle timeout on the engine side, but I did not confirm that and am not claiming it. What I can say precisely is the part on this side: **whatever ends the session, the agent does not recover from it.**

## Why it is worth fixing rather than documenting

The failure does not name its cause. A consumer sees eight `FAILED` statements referencing a session UUID they never created and cannot look up. In our case that landed one level worse: the calling step printed `FAILED` for each table and still returned 0, so the pipeline carried on and built gold against a catalog with nothing in it. That bug was ours and is fixed — but the reason it was expensive is that `session … is not running` reads like a transient blip rather than a permanent state, and the pipeline had no way to tell the difference.

Because the state is permanent per process, retrying at the caller cannot help. Only the agent can fix it.

## Shape of a fix

Catch the session-not-found condition at the point a statement is executed and rebuild the session before failing — i.e. treat `getOrCreate()` as something to re-run rather than an import-time constant. A bounded retry (rebuild once, re-execute, then surface the original error) would turn a permanent outage into a hiccup, and would compose with `connectconf.apply()`, which is already written to run per session precisely because the engine may have changed underneath.

## Versions

- `fabric-emulator-spark-agent` 0.29.0
- `fabric-emulator-sail` 0.29.0
- Consumer: `databricks-emulator` 0.2.4 over the same shared agent

calvinchengx:
Closes #312. The agent built its `SparkSession` at import and nothing ever ran
`getOrCreate()` again, so once the engine dropped that session every statement
failed with `session <uuid> is not running` until the container was restarted.

## The two decisions worth arguing about

**It does not re-run the statement.** A cell that wrote data and then met the
dead session on a later line would write twice on a silent retry, and the agent
cannot tell those halves apart — it hands `exec()` a block of arbitrary user
code, not a plan it can inspect for side effects. So recovery makes the **next**
statement work and lets this one fail, which is also what a real Livy session
does when its Spark application dies.

**It cannot be silent.** Temp views, cached DataFrames and session-scoped conf
lived in the engine session that went away. A "transparent" reconnect hands back
a notebook that has quietly forgotten its temp views — the same failure wearing
a friendlier face, and harder to diagnose than the error it replaces. The user
gets a line naming exactly what was lost and telling them to re-run.

## Shape

`session_recovery.py` holds the logic and imports no pyspark, so it is unit
testable — the same split as `catalog.py`, `jvmconf.py` and `connectconf.py`,
which exists because `agent.py` starts Spark at import. `agent.py` gains
`build_spark()` (the one place a shared session is created, so recovery can
re-run it) and `recover_lost_session()`.

**Every namespace is rebound, not just the module handle.** The per-Livy-session
objects come from `newSession` on the shared one, so they die with it; rebinding
only `spark` would leave every open notebook holding the corpse — the same bug
one level down. User variables are left alone: a plain Python value survives an
engine restart, and clearing namespaces would throw away work the engine never
held.

**Detection reads the message, not the exception type,** because the two engines
behind this one image disagree: Sail raises `IllegalArgumentException` with
`is not running`, classic Spark Connect raises `SparkConnectGrpcException` with
an `INVALID_HANDLE.*` error class. Matching on type would have recovered on one
engine and not the other, with nothing to say which.

## Verified

16 new tests; `pytest python/tests` 995 passed / 1 skipped; `make lint`,
`make check` and `go build ./...` all exit 0. The Python 3.8 floor test globs
agent sources, so the new module is covered by it automatically.

Mutation-tested rather than asserted — every one of these fails the suite:

| mutation | tests that catch it |
|---|---|
| every error looks like a lost session | 4 |
| `rebind` stops after the first namespace | 1 |
| the note stops naming what was lost | 1 |
| `annotate` writes only `evalue`, not the traceback | 1 |

## I deleted my own dead branch, having just complained about one

The first draft had a `NOT_LOST_MARKERS` deny-list so
`INVALID_HANDLE.SESSION_ALREADY_EXISTS` could not trigger a rebuild. Mutation
testing showed **deleting the whole guard failed no test** — none of the
positive markers matches that message anyway, so the branch could never change
an outcome. That is the same shape I reported in #326 (`isTextAnnotation`) and
declined to fix there, because there it would change behaviour; here it changes
nothing, so it goes. The test asserting `SESSION_ALREADY_EXISTS` is not treated
as a lost session stays and now guards the marker list against being widened to
something as loose as `"session"`.

## Not covered

That `do_POST` calls the recovery in the right place is not unit tested —
`agent.py` starts Spark at import, so no unit test can load it. The logic it
calls is tested; the call site is two lines. Exercising it end to end would mean
killing a live Sail session mid-suite, which is worth doing and is not this PR.

### DIFF
--- a/python/spark_agent/agent.py
+++ b/python/spark_agent/agent.py
@@ -27,16 +27,27 @@ from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 
 import httpjson
 import jvmconf  # no Spark; JVM session configs (see jvmconf.py)
+import session_recovery
 from pyspark.sql import SparkSession
 
-_b = SparkSession.builder.appName("livy-agent")
-if os.environ.get("SPARK_REMOTE"):
-    spark = _b.remote(os.environ["SPARK_REMOTE"]).getOrCreate()
-else:
+
+def build_spark():
+    """The one place a shared SparkSession is created, so recovery can re-run it.
+
+    Was inline at import until issue #312: when the engine dropped the session
+    there was no second caller of getOrCreate(), so the agent stayed broken
+    until the container restarted. Naming it changes nothing at boot.
+    """
+    b = SparkSession.builder.appName("livy-agent")
+    if os.environ.get("SPARK_REMOTE"):
+        return b.remote(os.environ["SPARK_REMOTE"]).getOrCreate()
     # Classic JVM: Delta + OneLake ABFS must be on the session that created
     # it. Without the extension, saveAsTable dies on
     # DELTA_CONFIGURE_SPARK_SESSION_WITH_EXTENSION_AND_CATALOG.
-    spark = jvmconf.configure(_b, os.environ).getOrCreate()
+    return jvmconf.configure(b, os.environ).getOrCreate()
+
+
+spark = build_spark()
 
 
 def _normalise_connect_confs():
@@ -320,6 +331,30 @@ def ns(session):
     return namespaces[session]
 
 
+def recover_lost_session():
+    """Rebuild the shared session after the engine dropped it, and rebind.
+
+    Returns the note to show the user, or None if the rebuild itself failed —
+    in which case the original error stands, because reporting a recovery that
+    did not happen is worse than the failure it replaces.
+    """
+    global spark
+    try:
+        spark = build_spark()
+    except Exception:  # noqa: BLE001 — engine still down; the caller reports the real error
+        print("agent: rebuilding the Spark session failed:\n" + traceback.format_exc(),
+              file=sys.stderr, flush=True)
+        return None
+    _normalise_connect_confs()
+    _install_delta_ops()
+    rebound = session_recovery.rebind(
+        namespaces, spark, catalog.isolate,
+        attach_sc=(None if os.environ.get("SPARK_REMOTE") is None else rddfacade.attach))
+    note = session_recovery.note(rebound)
+    print(note, file=sys.stderr, flush=True)
+    return note
+
+
 def remember_context(session, req):
     """Merge identity fields from a /mount or /statements body into this session."""
     cur = session_context.setdefault(session, {})
@@ -465,6 +500,15 @@ class Handler(BaseHTTPRequestHandler):
                             result = sqlrun.run_sql(req.get("code", ""), ns(session))
                         else:
                             result = run_code(req.get("code", ""), ns(session))
+                    # The engine dropping our session is not this statement's
+                    # fault and, until #312, was permanent: nothing re-ran
+                    # getOrCreate(), so every later statement failed the same
+                    # way. Rebuild now so the NEXT one works, and tell the user
+                    # what the rebuild cost. This statement still fails —
+                    # re-running arbitrary user code that may already have
+                    # written data is not something the agent can do safely.
+                    if session_recovery.envelope_is_lost_session(result):
+                        session_recovery.annotate(result, recover_lost_session())
             finally:
                 try:
                     import files_mount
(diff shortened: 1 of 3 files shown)


calvinchengx:
Pushed `6ebb810`: `is not running` alone was too loose a marker — it matches "the container is not running", "the Docker daemon is not running" and any engine-side message about something else having stopped, and every false positive would cost an open notebook its temp views. It now requires `session` beside it, which is Sail's actual text (`invalid argument: session <uuid> is not running`). Two tests added; loosening it back fails the suite.

On the red run before that push — **three jobs cancelled at 25m15s each and `Both silver builds agree, row for row` failing on `Artifact not found`**: that is a timeout signature, not a test failure, and the same pattern is on `main`, which cannot contain this branch. Main run [32215779213](https://github.com/calvinchengx/fabric-emulator/actions/runs/32215779213) has `Medallion — PySpark…` cancelled and the same comparison job failing. Runner contention is the likely common cause; I contributed to it earlier by dispatching ten cron workflows while the queue was already deep.

Re-running the failed jobs now, which also re-tests the tightened marker.

calvinchengx:
Root cause of the red, and it is not this branch — nor runner contention, which was my first guess and was wrong.

Every one of the four cancelled jobs spent its entire 25 minutes in one step:

```
cancelled  07:45:36 -> 08:10:45  Microsoft ODBC Driver 18 (pyodbc warmup) and mssql-python libs
skipped                          Run uv run --frozen --no-sync python e2e/medallion/run.py
```

The medallion run was **skipped** — this branch's code never executed. The stall is in fetching Microsoft's ODBC packages. The two `fail` jobs are the downstream comparisons reporting `Artifact not found` because their producers never got that far.

Same shape on `main`, which cannot contain this branch: its cancelled job stalled in `playwright install --with-deps chromium`. Two different jobs, two different package sources, same signature — external downloads hanging to the job timeout.

Re-running the failed jobs.

PR #329, opened 2026-08-19
ci: bound the package installs that can eat a whole job

calvinchengx:
Today `packages.microsoft.com` stalled and took **six jobs to their 25-minute
timeout** across two runs, on a PR whose code never executed:

```
cancelled  07:45:36 -> 08:10:45  Microsoft ODBC Driver 18 (pyodbc warmup) and mssql-python libs
skipped                          Run uv run --frozen --no-sync python e2e/medallion/run.py
```

Nothing in that picture names the mirror. The step has no budget and no retry,
so a stall consumes the whole job; `run.py` is reported `skipped`, and the
visible failure is the downstream comparison saying `Artifact not found for
name: silver-summary-medallion`. The run reads as a broken test.

`main` hit the same shape the same day from a different source — its cancelled
job stalled in `playwright install --with-deps chromium`. Two jobs, two package
repositories, one signature.

## What this changes

- **`timeout-minutes` on the three network-install steps** — 8 for ODBC, 10 for
  the two playwright installs. ~10x their normal time, so a stall fails fast and
  is attributed to the step that stalled, while the 25-minute job budget stays
  with the thing being tested. These are the first step-level timeouts in the
  file; the 63 existing ones are all job-level, which is exactly why a stalled
  step could spend the job's whole budget.
- **Retries on the network operations only.** `curl` gains `--max-time 120`
  (its default is *no* timeout, which is how one step ate 25 minutes) and
  `apt-get` gains `Acquire::Retries` and an HTTP timeout. `install` and `tee` are
  deliberately left unretried — a failure there is a real error, and retrying it
  would hide it three times.

## Verified

`bash -n` on the extracted step, and the retry helper exercised in all three
modes it has to get right:

```
A: first-try success -> ok
B: succeeded on attempt 2 -> ok        (one ::warning:: emitted)
C: gave up, non-zero -> ok             (three ::warning:: then ::error::)
D: set -e survived a failing first attempt -> ok
```

D is the one worth stating: under `set -euo pipefail` a naive retry aborts the
job on the first failed attempt instead of retrying. The helper calls the
command in a condition context, so it does not.

The workflow parses and the three step timeouts land where intended:

```
step timeout  10m  [portal] playwright install --with-deps chromium
step timeout  10m  [portal-terminal] playwright install --with-deps chromium
step timeout   8m  [medallion] Microsoft ODBC Driver 18 (pyodbc warmup) …
```

## What this does not do

It does not make a stalled mirror succeed. It makes the failure fast, retried
once or twice for free, and legible as what it is — rather than a 25-minute
silence ending in someone else's missing artifact.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1521,6 +1521,18 @@ jobs:
     timeout-minutes: 25
     steps:
       - uses: actions/checkout@v7
+      # The .debs, so a stalled mirror is a cache hit rather than a lost job.
+      # An OPTIMISATION, not a dependency — exactly like the spark-jars cache
+      # above: a cold or unusable cache falls through to the network path below,
+      # which is the one that was here before and is unchanged in substance.
+      #
+      # Scoped by runner.arch as well as OS because the .debs are architecture
+      # specific, and carrying a -v1 so a package-list change can invalidate it
+      # deliberately rather than by hoping the key changes.
+      - uses: actions/cache@v6
+        with:
+          path: ~/odbc-debs
+          key: odbc18-${{ runner.os }}-${{ runner.arch }}-msodbcsql18-unixodbc-krb5-v1
       - name: Microsoft ODBC Driver 18 (pyodbc warmup) and mssql-python libs
         # BOUNDED, because an unbounded install here does not fail — it eats the
         # job. 2026-08-19: packages.microsoft.com stalled and this step ran for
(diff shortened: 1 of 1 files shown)


calvinchengx:
Added the `.deb` cache (`101e46c`), so the mirror leaves the critical path.

`actions/cache` over `~/odbc-debs`, keyed on OS + arch + the package list. On a hit the step runs `dpkg -i` against the cached files and **contacts no network at all**. On a miss it runs the network path, then populates the cache — `apt-get clean` first so the archive holds exactly our dependency closure and nothing the runner image happened to leave behind.

**It is an optimisation, not a dependency** — the same shape as the `spark-jars` cache already in this file. If the cache is cold, or no longer satisfies the runner image (a dependency the image used to provide and no longer does would fail `dpkg -i`), it falls through to the network path and repopulates. A stale cache costs a slow job, never a red one.

Exercised all three paths against fakes, because the fallback is the part that must not be wrong:

```
A: cache HIT                  -> installed from 1 cached .debs; the mirror was not contacted
B: cache MISS                 -> ::notice::…cold or unusable; fetching from packages.microsoft.com
C: cache present, dpkg REJECTS-> ::notice::…cold or unusable; fetching…   (exit 0, not a failure)
```

Both paths end with `dpkg -s msodbcsql18` and `odbcinst -q -d`, so neither is trusted on its say-so — a half-installed driver would otherwise surface much later as a confusing pyodbc error inside the medallion run.

**One honest limit:** GitHub scopes caches by branch. A cache written on this PR branch is not visible to other PR branches, so the benefit starts once this lands on `main` and the first main run populates it. This PR's own CI still depends on the mirror, and may still be red on an unlucky runner — which is the situation the timeout half already makes fast and legible rather than a 25-minute silence.

PR #332, opened 2026-08-19
build(deps): kafka-python 3.0.11

calvinchengx:
One quarter of #319, which bundled four majors and a wholesale re-resolution.

`kafka-python` 2.3.2 → **3.0.11**, moving **exactly one package**.

Widening the bound alone is a no-op (`uv lock` keeps 2.3.2), so this widens it
*and* runs `uv lock --upgrade-package kafka-python` — the smallest change that
delivers the version.

Local: `pytest python/tests` 997 passed / 1 skipped.

The suite that judges this is `Eventstream notebook API on Sail + Kafka`, which
failed in #319. A major bump of the Kafka client is exactly the kind of thing
that should be judged on its own rather than alongside a numpy jump.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -168,7 +168,7 @@ sail-delta = [
     "requests>=2.32,<3",
     # OSS format("kafka") on Sail: the agent consumes and createDataFrames
     # Kafka-schema rows into the engine. JVM keeps spark-sql-kafka.
-    "kafka-python>=2.0.2,<3",
+    "kafka-python>=2.0.2,<4",
     # JKS/P12 → PEM so kafka-python can honour Spark's Java truststore options.
     "pyjks>=20,<22",
     "cryptography>=42,<50",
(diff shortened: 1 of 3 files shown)


calvinchengx:
Pushed the fix. kafka-python 3 does not ignore `api_version_auto_timeout_ms` — it raises:

```
kafka.errors.KafkaConfigurationError: Unrecognized configs: {'api_version_auto_timeout_ms'}
```

so every eventstream read failed before it connected.

The config bounds the broker version probe, and it is now passed **only when the installed client declares it**, rather than being deleted or gated on a version number:

```python
if "api_version_auto_timeout_ms" in getattr(KafkaConsumer, "DEFAULT_CONFIG", {}):
    conf["api_version_auto_timeout_ms"] = min(timeout_ms, 10000)
```

Asked of the client rather than pinned, because this agent image is consumed by more than one emulator and they do not upgrade in step — the same reasoning `connectconf.py` already applies to the Sail byte-size conf.

Executed against fakes of both majors rather than reasoned about:

```
kafka-python 2 (has the config)    -> passes it: True
kafka-python 3 (removed it)        -> passes it: False
```

`pytest python/tests` 997 passed / 1 skipped, `make lint` clean.

PR #333, opened 2026-08-19
build(deps): mcp 2.0.0

calvinchengx:
One quarter of #319, which bundled four majors and a wholesale re-resolution.

`mcp` 1.29.0 → **2.0.0**, which brings **five packages with it**:

```
mcp        1.29.0 -> 2.0.0
mcp-types  (new)  -> 2.0.0
httpx2     (new)  -> 2.12.0
httpcore2  (new)  -> 2.12.0
httpx2-jsfetch (new) -> 1.0
```

A new HTTP stack arriving alongside the SDK is worth seeing on its own — in
#319 it was one of 25 moving packages and there was no way to attribute the
`Python mcp SDK witnesses Fabric Core MCP` failure to it rather than to
anything else.

Widening `mcp>=1.12,<2` to `<3` is a no-op by itself (`uv lock` keeps 1.29.0),
so this widens the bound *and* runs `uv lock --upgrade-package mcp`.

Local: `pytest python/tests` 997 passed / 1 skipped. The witness that judges it
is `ci:mcp-core`, which drives the real Python MCP SDK against the emulator.

### DIFF
--- a/e2e/mcp-core/driver.py
+++ b/e2e/mcp-core/driver.py
@@ -19,6 +19,7 @@ import sys
 import urllib.parse
 import urllib.request
 
+import httpx2
 from mcp import ClientSession
 from mcp.client.streamable_http import streamable_http_client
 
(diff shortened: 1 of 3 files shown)


calvinchengx:
Pushed the fix. mcp 2.0 renamed `streamablehttp_client` to `streamable_http_client` — the error even suggested it:

```
ImportError: cannot import name 'streamablehttp_client' from 'mcp.client.streamable_http'. Did you mean: 'streamable_http_client'?
```

The driver now uses the 2.x spelling, and **the bound floor moves with it** — `mcp>=2,<3` rather than `>=1.12,<3`. Leaving the floor at 1.12 would let a fresh resolution pick a client this import cannot find, and the suite would then fail on `ImportError` rather than on anything about the emulator. The reason lives beside the pin in `pyproject.toml`, since that is where someone would otherwise widen it back.

`pytest python/tests` 997 passed / 1 skipped, `make lint` clean.

calvinchengx:
Second break in the same bump, and it needed the signature rather than the symbol:

```
TypeError: streamable_http_client() got an unexpected keyword argument 'headers'
```

mcp 2.0's signature is now:

```python
streamable_http_client(url: str, http_client: httpx2.AsyncClient | None = None, terminate_on_close: bool = True)
```

The per-call `headers`/`auth`/`timeout` kwargs are gone; the caller supplies a configured client. So the bearer and the emulator's self-signed cert are now set once on an `httpx2.AsyncClient` instead of per call.

**`httpx2`, not `httpx`** — that is the new stack this bump drags in (`httpx2`, `httpcore2`, `httpx2-jsfetch`, `mcp-types`), and passing an ordinary `httpx.AsyncClient` would fail on the type rather than on the request. Worth knowing for anything else in the repo that later talks to this SDK.

This is the second failure #319 could not have attributed: bundled with 24 other package moves, `Python mcp SDK witnesses Fabric Core MCP` going red said nothing about which of them caused it.

`make lint` clean, `pytest python/tests` 997 passed / 1 skipped.

PR #334, opened 2026-08-19
build(deps): great-expectations 1.20.0 — and numpy 2

calvinchengx:
One quarter of #319, and **the one carrying the real risk**.

`great-expectations` 0.18.22 → **1.20.0** does not come alone:

```
great-expectations  0.18.22 -> 1.20.0
altair              4.2.2   -> 6.2.2
numpy               1.26.4  -> 2.5.2      <- the whole point of splitting
```

**numpy 2 is the change here**, not Great Expectations. It sits under pandas,
pyarrow, Spark's Python side and every notebook the emulator runs, so its blast
radius is the entire Python surface. In #319 it arrived as one of 25 moving
packages, which is why three Great-Expectations jobs failing said nothing about
whether GE 1.x or numpy 2 was responsible.

Splitting isolates it: if this PR is red and the other three are green, numpy 2
is the answer. That is the question #319 could not ask.

Also note GE 0.18 → 1.x is a documented rewrite of the API surface
(`great_expectations.dataset` → the new fluent/context API), so the suite may
need source changes rather than just a bump. If so, this PR should become that
work rather than a version line.

Widening `great-expectations>=0.18,<1` to `<2` is a no-op by itself, so this
widens the bound *and* runs `uv lock --upgrade-package great-expectations`.

Local: `pytest python/tests` 997 passed / 1 skipped — which proves only that
*our own* Python tests survive numpy 2, not that the GE suite does. The three
`Real Great Expectations validates semantic-model data` jobs are the judge.

### DIFF
--- a/e2e/great-expectations/driver.py
+++ b/e2e/great-expectations/driver.py
@@ -48,6 +48,11 @@ def query(dax):
     return df
 
 
+# ty reports every keyword below as `unknown-argument`: GE 1.x builds its
+# Expectation classes from pydantic models, so a static checker sees no declared
+# parameters on them. The keywords are correct — the offline harness and CI both
+# execute these calls — so the suppressions are about the checker's view of a
+# third-party library, not about the calls.
 def batch(df):
     """A GE 1.x Batch over `df`, ready to validate single Expectations.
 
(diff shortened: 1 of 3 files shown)


calvinchengx:
Ported the driver to the 1.x API (`c7eeaed`, `+ lint fix`). The shape survives; the spelling does not.

```
0.18                                     1.x
ctx.sources.add_pandas(...)              ctx.data_sources.add_pandas(...)
asset.build_batch_request(dataframe=df)  asset.add_batch_definition_whole_dataframe(...)
ctx.get_validator(batch_request=...)     batch_def.get_batch(batch_parameters={"dataframe": df})
v.expect_column_values_to_be_between(…)  batch.validate(gxe.ExpectColumnValuesToBeBetween(…))
```

One batch per asset and one result per expectation, exactly as before, because `validate()` returns the same `ExpectationValidationResult` — `.success` and `partial_unexpected_list` both still there, so the tutorial's assertions are untouched.

**Verified offline, without the emulator**, by stubbing `executeQueries` and running the real driver through `runpy`:

```
PASS  Retail Store: row_count_between(1,10)
PASS  Retail Store: valid_zip5(PostalCode)
PASS  Retail Measure: TotalUnits >= 50000
FAIL  Retail DAX: Total Units Ratio in [0.8, 1.5]
GREAT-EXPECTATIONS E2E: PASS (Store/Measure pass; YoY ratio fails as in the tutorial)
```

That the harness can fail was checked too: changing the stubbed 1.8 to 1.2 makes it exit 1 on `AssertionError: YoY ratio should fail`. Otherwise a green offline run would only prove the driver imports.

**Four `ty: ignore[unknown-argument]` suppressions**, with the reason recorded once above `batch()`: GE 1.x builds its Expectation classes from pydantic models, so a static checker sees no declared parameters and reads every keyword as unknown. The keywords are right — these calls execute offline and in CI. Worth noting the directive has to sit on the ARGUMENT line of a wrapped call, not the line that opens it; ty anchors the diagnostic there.

`make lint` clean, `pytest python/tests` 997 passed / 1 skipped. The three `Real Great Expectations validates semantic-model data` jobs — and whatever numpy 2 does to the rest — are the real judges.

PR #342, opened 2026-08-19
docs: gate doc links as well as sidebar reachability

calvinchengx:
Adds `scripts/check_docs_links.py --strict` and wires it into CI, so the family's docs are gated the same way everywhere.

## Why the site build is not enough

Measured, not assumed: the Astro build **exits 0** on a link to a page that does not exist, and publishes a 404. It also cannot see a page missing from the sidebar, which still builds, still gets a URL and still turns up in search, while being unreachable by navigation.

## What it checks

- every intra-doc markdown link resolves
- every README link into `docs/` resolves
- every sidebar `slug:` has a page behind it
- every published page appears in the sidebar

**The published set is derived, not copied.** The checker reads `DOC_RE` out of `website/scripts/sync-docs.mjs`, so it guards exactly the set the site publishes. A second copy kept in step by a comment is a defect already filed, with no owner and no failing test. If that declaration ever moves, the checker fails loudly instead of silently guarding the wrong set.

Sidebar entries under `parity-history/` are exempt, because `parity-versions.mjs` generates those from git tags rather than from `docs/`. The exemption applies only when that generator is present, so a repo without it still has a dangling slug reported.

## Verification

Every failure mode was confirmed by deliberately introducing it, then reverting: a broken link, an orphan page, a sidebar entry with no page, and a moved `DOC_RE`. All four exit 1; the clean tree exits 0. The workflow YAML was parsed after editing.

## Overlap with `check_docs_sidebar.py`

This repo already guards the sidebar half, with its own tests. The new checker
covers that half too, so they overlap deliberately: both pass or both fail, and
retiring the older one is its owner's call rather than a side effect of adding
link checking. The docstring says so.

This repo passes cleanly as it stands: 54 published pages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -462,6 +462,10 @@ jobs:
       # out still builds and still has a URL, but nothing links to it, so it is
       # published and unreachable. Same drift, same guard.
       - run: uv run --frozen --no-sync python scripts/check_docs_sidebar.py
+      # The site build does NOT catch a dangling intra-doc link: measured, it
+      # exits 0 and publishes a 404. Nor can it see a page missing from the
+      # sidebar, which is published but unreachable by navigation.
+      - run: uv run --frozen --no-sync python scripts/check_docs_links.py --strict
       # THE ONE `make check` RAN AND CI DID NOT. Every example must resolve its
       # target through the `fabric-target` contract — no pinned seed principal,
       # no localhost endpoint outside the resolver, no emulator-only lever, and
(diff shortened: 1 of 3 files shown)


calvinchengx:
Pushed tests. The two failing jobs were mine: this repo measures coverage over `scripts/` with `fail_under = 92`, so adding an untested ~85-statement checker dropped the suite below the floor. That is the gate working, not a flake.

`python/tests/test_check_docs_links.py` now covers all four failure modes plus the derivation itself, in the same shape as `test_check_docs_sidebar.py`. Locally: **1029 passed, coverage 93.47% against the 92 floor**.

PR #364, opened 2026-08-23
docs site: a landing page at the root, and a docs lane in CI

calvinchengx:
#363 was one line in a markdown generator. It ran **95 checks** on the PR and the same matrix again on main: three OSes of Go tests, every compose-backed conformance suite, dbt, Terraform, the medallion pairs. None of them can observe a markdown file.

**The lane.** `scripts/docs_only_change.py` classifies the diff; every job in `ci.yml` except `witnesses` is gated on it. `witnesses` is deliberately ungated: it is the job that reads the docs (parity claims and witnesses, sidebar, intra-doc links, conformance matrix), so it runs either way. The lane removes the jobs that cannot see the change, not the ones that check it.

**Three things that are load-bearing:**

- *Unknown means run everything.* Empty diff, null `before` on a force push, an unresolvable range: all answer `false`. A classifier that guesses "docs" turns a missing verdict into a green one.
- *A deletion or rename is never docs-only.* `internal/api/examples_readme_test.go` reads `examples/README.md` and fails when a `(../docs/<file>)` link points at nothing, so removing a page can break a Go test. Found by looking, not assumed.
- *Skipped, not absent.* Job-level `if:`, not `paths-ignore:` on the workflow, so a docs PR shows sixty-odd jobs explicitly declining and the classifier's reasoning is in the run log. A path-filtered workflow is missing from the rollup entirely, and green-because-it-did-not-run is the false all-clear this repo keeps finding.

**Two consequences handled rather than discovered:**

- The badge step took the newest successful CI run on main. With the lane, that run may have skipped `coverage-e2e`, so every typo fix would have replaced real coverage numbers with `n/a`. It now walks back until a run actually yields the artifact.
- `docs-site.yml` now builds on docs PRs, which nothing did before (an Astro failure was found after merge). It uses a **separate concurrency group**: sharing `pages` would put PR builds in the same single-slot queue as a real deploy, and GitHub keeps one pending run per group, so a burst would silently drop a publish.

**Not covered:** `make-targets.yml` (three jobs) still runs on everything. It is also the release gate via `workflow_call`, and a gate that can be skipped is worth more care than three runner-minutes.

Self-test runs beside the real classification on every run (12 cases, both directions), because `docs_only=false` is always safe and always plausible: a rule that stopped matching anything would be silently useless.


calvinchengx:
## Second commit: the landing page

I had answered the wrong question earlier. `data-agent-service` and `data-agent-voice` publish a **hand-built landing page at the site root**, with Starlight mounted underneath; what I had improved was fabric-emulator's Starlight index prose. This adds the real thing.

`website/src/pages/index.astro` wins over Starlight's catch-all, so `/fabric-emulator/` is now a landing page: hero and release pill, *the gap this is built around*, an architecture diagram, what it runs as status-tagged cards, **how the claims are checked** (the witness-kind ranking), an honest-status table of the 🟠/🔴 rows, get it, run it. The docs overview it replaced moves to `/overview/`, first in the sidebar.

**No published URL moved.** The docs were already served flat from the base, so only the root page changed. That is the one deliberate divergence from the siblings, which mount docs at `/docs/`: matching them would have moved 56 live URLs to save nothing.

**Every number is counted at build time**, none typed. `sync-docs.mjs` writes `site-stats.json` from `docs/parity.md` and `docs/witnesses.json`: 138 of 158 graded capabilities real, 599 witnesses over 126 claims, 45 CI jobs, 63 e2e suites, 56 pages. The page this replaces hardcoded "113 supported capability claims" and was wrong by seven within a fortnight.

### A bug found on the way, visible on the published site

`PARITY_RE` was `/(^|[/-])parity\.md$/`, which also matches **`29-tsql-parity.md`** — and since both `ls-tree` and `readdirSync` return sorted names, the T-SQL page won every lookup from the day it was added. So the live parity map was read from the wrong file, and the published changelog has been saying:

> **Current (latest-…):** 0 🟢 Real · 0 🟡 Emulated · 0 🟠 BYO-engine · 0 🔴 Not implemented

because that page grades with ✅/❌. The version picker's "latest" pointed at `/29-tsql-parity/` too. Nothing failed: a generator that finds the wrong file still produces a page. Now matched exactly, and the changelog reads 138 / 9 / 3 / 8.

Also derived rather than hardcoded: `check_docs_links.py` took its sidebar exemption from a literal `{"index"}`, which would have silently exempted a page that no longer exists while reporting one that was fine. It now reads the set out of the generator, and errors if that matches nothing.

PR #372, opened 2026-08-23
docs: restore the 105 published URLs the base move 404s, and preview the real site

calvinchengx:
Two things, because the second is what found the first.

## The base move 404s every URL this site ever published

#368 moved the docs from `/fabric-emulator/<slug>/` to `/fabric-emulator/docs/<slug>/`. All **105** routes the site published before that are 404 on the live site right now, and three of them are still linked from committed documentation in this family:

```
calvinchengx.github.io/fabric-emulator/10-testing/     -> 404   (3 references)
calvinchengx.github.io/fabric-emulator/12-e2e-matrix/  -> 404   (1 reference)
calvinchengx.github.io/fabric-emulator/parity/         -> 404   (1 reference)
```

`website/published-routes.txt` is that route list, captured by building the docs at `2d6e6de`, the commit before the base changed. It is the oracle: `scripts/assemble_site.py` writes a redirect stub for every entry and **fails the deploy** if any would 404. Deriving the check from the new build instead would only prove the new build agrees with itself.

Astro's `redirects:` cannot do this: a redirect key is emitted underneath the configured base, so `/parity/` would publish at `/fabric-emulator/docs/parity/` and nothing would answer the root-level path anyone linked to.

This replaces the two `cp` lines in the workflow. Ported from azure-keyvault-emulator, entra-emulator and databricks-emulator, which already had it.

## One target builds and serves the site

```
make docs-build    # build ./_site: what CI deploys
make docs-serve    # ...and serve it at the URLs it publishes under
```

`astro dev` cannot show this site: it is based at `/fabric-emulator/docs/` and knows nothing about the tree around it, so under it the landing page does not exist, the stubs do not exist, and the badge documents the landing page fetches do not exist. The preview server routes the way Pages does, and `resolve()` is a pure function with self-test cases.

**CI now runs `make docs-build`.** The one thing it cannot reproduce locally is the coverage numbers, which come from the last CI run's artifact; the target writes "n/a" badges in their place, the same fallback the workflow already uses when no artifact is found, and the workflow copies the real ones over them.

### The favicon 404s on every page

Starlight defaults `<link rel="icon">` to `/favicon.svg` and does not check the file exists. Every one of the eight family sites has this, measured against the live sites. Fixed, and guarded: the assembler resolves every absolute `href`/`src` in every built page and fails if the pattern stops matching at all. 8485 references check today.


calvinchengx:
The failure here is not yours. Main is red on two counts and this PR inherits both:

1. **Four tests broken by #367**, which added the top-level-`permissions:` invariant to `check_workflow_concurrency.py` without updating the test fixtures. The two workflows it names, `ci.yml` and `make-targets.yml`, are the synthetic ones the tests write into `tmp_path`; the real workflows all declare permissions. Fixed by [#373](https://github.com/calvinchengx/fabric-emulator/pull/373).
2. **The coverage floor**: 92% required, main is at 90.51% because two scripts merged with no tests. Fixed by [#374](https://github.com/calvinchengx/fabric-emulator/pull/374), which also carries #373's commit so it can be green.

Rebase once #374 lands and this should go green untouched.

calvinchengx:
Pushed two commits here to clear CI (this is another session's PR, so: merge of main, plus tests, nothing touched in the change itself except one guard noted below).

**1. Merged main.** The four `test_check_workflow_concurrency` failures were #367's stale fixtures, fixed on main in [#374](https://github.com/calvinchengx/fabric-emulator/pull/374).

**2. The coverage floor, again.** With main merged the suite passed but the total was **90.91%** against `fail_under = 92`, because `scripts/assemble_site.py` (159 statements) arrived with no pytest. Now **93.89%**, 1246 passed.

The tests are built around what this script actually guarantees, which is negative: *nothing that used to resolve stops resolving*. That fails silent by construction. If `routes_in` stops finding pages, no stub is written, the oracle compares against an empty set, and the run is green while 105 URLs are dead. So each check is driven from the failing side first, with the passing case as the control:

- a route in the oracle that the build no longer emits, which must be named and must fail;
- an empty and a missing `published-routes.txt`, both refused rather than passed over;
- a reference to a file never shipped (the favicon case), a reference outside the project prefix, and a page with **no** absolute references at all, which must fail rather than pass;
- the `ALIASES` branch, which is empty in the repo today so nothing else exercises it;
- `resolve` across all nine routing cases, because a preview that answers requests the real host refuses is worse than no preview;
- the preview server itself over a real socket, asserting the 302 at `/`, the 404 outside `/fabric-emulator/`, and a pre-move URL being answered by its stub;
- a control that neuters `write_stub` and asserts the script's own `--self-test` goes red.

**One change to `assemble_site.py`**, two lines: `self_test()` read `kept/index.html` right after asserting it might not exist, so a self-test that had *already caught* the bug died with a traceback instead of printing its four lines and returning 1. Now guarded with `and have`. That control test is what surfaced it.

Verified end to end on this branch, not just in unit tests: `pnpm --filter fabric-emulator-docs build` then `scripts/assemble_site.py --out _site` gives **105 pre-move routes still resolving** and **8485 absolute references resolving**.

calvinchengx:
⚠️ **Re-applied after a force push.** The rebase to `0b53183` dropped `083d0f1` (`python/tests/test_assemble_site.py` plus a two-line guard in `assemble_site.py`). I have put it back on top of your new head, unchanged.

Please keep it on any further rebase: **without it this PR cannot go green.** `scripts/assemble_site.py` is 159 statements at 0% coverage, which puts the total at 90.91% against `fail_under = 92`. With the tests it is 93.87%, 1246 passed, ruff clean.

If you would rather write the tests yourself, say so and I will drop mine, but please do not simply rebase them away again.

calvinchengx:
**Green: 100 checks, zero failures**, at `bb2d0d7`.

Three commits from me on top of yours, all additive:

1. `test_assemble_site.py` — 36 tests, coverage 90.91% → 93.70%. That was the blocker: `assemble_site.py` is 159 statements and arrived at 0%, under the 92% floor.
2. A two-line guard in `assemble_site.py`'s `self_test()`: it read `kept/index.html` immediately after the check that finds it missing, so a self-test that had *already caught* the bug died with a traceback instead of printing its four lines and returning 1.
3. The preview-server test no longer drives a real socket. Worth stating plainly: my first two versions passed locally and **failed twice on the macOS runner**, the thread never binding within 30s with no exception to explain it. I did not find the cause, so rather than guess a third time I removed the socket. The routing decision is a pure function with nine cases already, and what remained — that the handler is wired to it — is checked without a listener. A test that reports a timeout for an unknown reason teaches people to press re-run.

Verified beyond the unit tests: `pnpm --filter fabric-emulator-docs build` then `assemble_site.py --out _site` gives **105 pre-move routes still resolving** and **8485 absolute references resolving**.

Yours to merge.

PR #373, opened 2026-08-23
fix: a failure message must not raise while reporting the failure

calvinchengx:
Rebased onto #374 and cut down to what that PR does not do. Everything else this branch carried (the permissions invariant test, the coverage for both scripts) landed there first, so it is gone from here.

## The defect

`build_landing_data.py` renders every failure message with `.relative_to(ROOT)`, which **raises** for a path outside the repo. Every caller is inside an error branch, so the branch whose whole job is to explain a problem replaced the explanation with a `ValueError` from pathlib.

It has never fired in production, where these are module constants under `ROOT`. It has also never been exercised, and #374's fixture says why:

```python
# Only used to shorten paths in messages; tmp_path is under it now.
monkeypatch.setattr(c, "ROOT", tmp_path)
```

Moving `ROOT` above the fixtures is the workaround for exactly this. With it removed, 11 of #374's own tests fail against the unfixed script, which is the measure of how much was being papered over.

## The fix

`shown()`, with two deliberate choices beyond the fallback:

- **`as_posix()`, not `str()`.** The three sibling checkers already do this. On Windows `str()` renders `docs\witnesses.json`, so one failure would read two ways across the three platforms this suite runs on. I know because I wrote `str()` first and only Windows CI caught it.
- **`PurePath`, not `Path`.** Nothing in it touches the filesystem, and the wider type is what lets the Windows rendering be driven from any platform, rather than needing a Windows runner to see a regression.

## Also: one uncovered guard

`engine_matrix` skips a row whose width disagrees with the header. That line was uncovered, and it is load-bearing rather than defensive: `zip(columns, cells[1:], strict=True)` **raises**, so without it one malformed line in a generated matrix takes down the docs deploy instead of publishing. Removing the guard now fails a test.

## Verification

```
1216 passed, 1 skipped     coverage 93.76%    build_landing_data.py 99.21%
ruff: All checks passed!   ty: All checks passed!
```

Three controls, each run with `__pycache__` cleared between steps:

| mutation | result |
|---|---|
| fallback removed | 11 tests fail |
| `as_posix()` back to `str()` | 1 test fails, on macOS |
| width guard removed | 1 test fails |

### DIFF
--- a/python/tests/test_build_landing_data.py
+++ b/python/tests/test_build_landing_data.py
@@ -114,10 +114,45 @@ def sources(tmp_path, monkeypatch):
         if write_stats:
             site_stats.write_text(json.dumps(stats), encoding="utf-8")
         monkeypatch.setattr(c, "SITE_STATS", site_stats)
-        # Only used to shorten paths in messages; tmp_path is under it now.
-        monkeypatch.setattr(c, "ROOT", tmp_path)
+        # ROOT is deliberately left alone. tmp_path is outside the repo, so
+        # every failure message these provoke names a path ROOT does not
+        # contain, which is the case `shown()` exists to survive.
 
     return build
+
+
+# --- rendering a path inside a failure message -----------------------------
+
+def test_a_path_inside_the_repo_is_shortened():
+    assert c.shown(c.ROOT / "docs" / "witnesses.json") == "docs/witnesses.json"
+
+
+def test_a_path_outside_the_repo_is_shown_whole(tmp_path):
+    # `Path.relative_to` raises for anything outside ROOT, and every caller is
+    # inside a failure message. Without the fallback the branch whose job is to
+    # explain a problem replaces the explanation with a pathlib traceback.
+    outside = tmp_path / "witnesses.json"
+    assert c.shown(outside) == outside.as_posix()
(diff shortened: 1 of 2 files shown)


calvinchengx:
Diagnosis from reviewing this and #372 together: the four failing tests here are fixed by this PR (16 pass at `2328a94` locally), but the job still fails on the **coverage floor** — 90.51% against `fail_under = 92`, because `scripts/build_landing_data.py` (121 statements) and `scripts/docs_only_change.py` (45) both merged with no pytest. So this cannot go green on its own.

[#374](https://github.com/calvinchengx/fabric-emulator/pull/374) covers both scripts (total 93.76%) and, to be green itself, **cherry-picks `2328a94` from here verbatim**. If this merges first, #374 rebases and drops the duplicate automatically. Happy either way — I did not want to touch your branch.

calvinchengx:
[#374](https://github.com/calvinchengx/fabric-emulator/pull/374) merged as `f13dec4` with all 100 checks green, and it carries your `2328a94` verbatim. So this PR's change is already on main and this can be closed without losing anything — leaving it to you rather than closing someone else's PR.

PR #479, opened 2026-09-15
Docs pointed at files that no longer exist; a checker now says so

calvinchengx:
Prose is the one artifact in this repo with nothing underneath it. A renamed
directory breaks a Go import and CI goes red; the same rename inside a sentence
breaks nothing at all, so the sentence keeps its confident tone and quietly
starts lying. Three had already drifted, and none is a typo — each is a fact
that expired, and the only thing that would ever have caught them is somebody
happening to click.

## The drift found, and fixed

| Doc | Cited | Reality | Fix |
|---|---|---|---|
| `docs/31:8` | `examples/medallion-pyspark/common.py` | the fixtures were extracted; the file is `examples/contoso-fixtures/common.py` | repointed |
| `docs/38:703` | `e2e/warehouse-tds` as a directory | the reader **is real and runs on every push** — it is a `go test` job, and no such directory has ever existed | named as the job it is |
| `docs/13:109` | `internal/store/fabric.go` as one of ours | a **real file in the SIBLING entra-emulator repo** | prose now says whose it is |

The third is worth a second look. The plan for this work assumed it was stale
and should be repointed at whichever local file holds the
`Active/Provisioning/Failed` state object. It isn't stale — `../entra-emulator/internal/store/fabric.go`
exists, and the paragraph is correctly describing entra-emulator's
implementation. Repointing it would have replaced a true sentence with a false
one. The honest fix is to say whose file it is, and exempt the token because
this checker cannot resolve a path across repos.

## What the checker asserts

`scripts/check_doc_drift.py`, stdlib-only, in the sibling `check_*` family —
wired into `make check` beside `check_docs_sidebar.py` and into the `witnesses`
CI job beside `check_docs_links.py`.

1. **Dead repo paths** — a backticked `dir/path` under a tracked top-level
   directory that is not on disk. The class with live drift.
2. **Dead make targets** — `make <target>` naming a target the Makefile does not
   define. Currently clean; a regression guard.
3. **Unread env vars** — a backticked `FABRIC_`/`ENTRA_`-style name no
   non-Markdown file reads. Currently clean; a regression guard.

Classes 2 and 3 being clean is the thing most likely to be a lie, so neither is
trusted on the strength of a green tree: both are tested against synthetic
drift, and both were measured to be non-vacuous against the real tree — class 2
extracts **124 command invocations across 20 distinct real targets**, class 3
extracts **53 env-shaped tokens** against **132 names found in code**. A checker
that always passes is indistinguishable from one that is working, which is
`docs/10`'s item Eight.

## Precision over recall, and the measurements behind it

A checker that cries wolf gets muted, and a muted check is item Eight again. The
first run produced **11 path hits, 9 of them false**, and each concession below
is a measurement rather than a preference:

- **`make` is anchored to command-shaped spans.** The naive `make \w+` regex was
  run against this tree: **13 hits, all 13 English prose** — "make the", "make
  it", "make every". Anchoring to an inline span or fenced line that reads like
  a command gives 0 false positives.
- **A dot that is not a known file extension is not a path**, so the Go symbol
  `internal/tsql.DataFlows` is left alone while `internal/api/livy.go` is not.
- **`docs/24` is this repo's shorthand for a numbered document**, not a path
  ("``docs/24``'s L is unchanged"). This was **not anticipated in the plan** and
  was the entire false-positive set of the first run — 9 hits across 5 files. No
  tracked file anywhere has a purely numeric name, so rejecting them costs no
  coverage.
- **Release notes are skipped.** A v0.16 note naming a since-renamed file is
  *correct* about the tree at that tag; editing it would falsify a historical
  record.
- **The env scan excludes by extension, never by directory.** `docs/demo/flow.py`
  and `docs/demo/flow-override.yml` are real code living beside prose — skipping
  `docs/` wholesale wrongly reports `DEMO_FABRIC_PORT` as read by nothing.

## The two exemptions, and why they are not edits

Both are forward references — prose that is *true about intent*. The alternative
to exempting them is editing accurate prose into vaguer prose to satisfy a tool.

- `docs/30` → `scripts/check_contracts.py`: the **O2 row of that document's own
  milestone table**, which describes work to come.
- `docs/54` → `e2e/task-parameters`: cited as a suite whose discipline is "worth
  having" — an argument for writing it, not a claim it exists.

(The `docs/13` cross-repo entry above is the third, and is a different category:
the path resolves, just not in this repo.)

Each entry carries a written reason, and `test_every_exemption_is_still_live`
fails if an exemption's document stops mentioning the token, or if the path
starts existing — so the hatch cannot quietly widen into a silencer.

## Verification

- `python3 scripts/check_doc_drift.py --strict` → exit 0 on the fixed tree.
- **Non-vacuity proven in place**: appending a dead path, a dead `make` target
  and an unread env var to a real doc makes all three classes fire and `--strict`
  exit 1; reverting returns it to 0.
- `ruff check` clean on both new files; `make check` runs `ruff + ty` and every
  sibling invariant — **all pass, nothing regressed**.
- `pytest python/tests` → **1823 passed, 1 skipped**. The new file is at
  **98.05%** coverage, nudging the repo total from 92.19% to 92.23% against its
  92.0% floor rather than spending the headroom.
- `test_a_variable_override_is_not_a_target` **caught a real bug while being
  written**: the first extractor captured `[A-Za-z0-9_.-]+`, which stopped *at*
  the `=` in `make PROFILE=lean` and handed the guard a plausible-looking
  "PROFILE" to flag. The word is now captured whole so the guard can see it.

Documented in one place — `docs/10-testing.md`, as item Twelve in the catalogue
of failures this codebase keeps producing, where the surrounding eleven are code
lying about code and this one is documentation doing it.

Nightshift-Task: doc-drift
Nightshift-Ref: https://github.com/marcus/nightshift


calvinchengx:
## Review fix — the blocking defect, and the three minors

**Blocker: `_TARGET_NAME` admitted a leading hyphen.** Confirmed by execution before changing anything:

```
`make -j4`             -> [(1, 'make -j4',  '-j4')]   # flagged
`make -C portal build` -> [(1, 'make -C',   '-C')]    # flagged
`make PROFILE=lean`    -> []                          # correctly skipped
```

Only the `PROFILE=lean` half of the guard's comment was ever true. Fixed by splitting the class — `^[A-Za-z0-9_.][A-Za-z0-9_.-]*$` — so a hyphen is legal inside a name and never at the front. Flags rejected; `check`, `up-jvm`, `docs-build` and `.PHONY` still accepted.

The green tree was the problem, not the evidence: no doc writes a make flag today, so nothing would have caught this until the first doc that did — and since this gates `make check` and the `witnesses` job, that edit would have hard-failed the build with "invokes a target the Makefile does not define: make -j4".

### The three minors, all fixed

| Finding | Fix | Verified |
|---|---|---|
| multi-target rules (`foo bar:`) unparsed | capture the whole LHS and split | real Makefile parses to the **same 23 targets** before and after; no `include` directives exist |
| `.exists()` is case-insensitive on macOS, sensitive on CI | decided against the **git index** (files + derived ancestor dirs) | checker still exits 0 on the tree, so no doc cites a gitignored path |
| ``` and ~~~ toggled interchangeably | a fence closes only on the marker that opened it | — |

The case fix is the one with real behaviour change, so I checked it does not over-flag: still `64 docs, 3 recorded forward references, no drift`.

### Non-vacuity, per test

Each new test was proven by reverting **its own** fix and watching it fail:

```
reverted case-exact lookup   -> FAILED test_a_wrong_case_path_fails_on_every_platform
reverted multi-target rules  -> FAILED test_a_rule_naming_several_targets_defines_all_of_them
reverted fence pairing       -> FAILED test_a_tilde_fence_is_not_closed_by_a_backtick_fence
reverted _TARGET_NAME        -> FAILED test_a_flag_is_not_a_target
```

That pass earned its keep: the fence test **passed with the bug still in place** on the first attempt, because it quoted ``` mid-line where the anchored `_FENCE` never matches. Rewritten to open the nested fence at a line start, where it actually inverts the state. Worth flagging as its own instance of the thing this PR is about — a test whose comment claimed a guarantee it was not exercising.

Re-ran the end-to-end probe against the real tree too, since `dead_paths` changed substantially; all four classes fire and revert clean:

```
dead path          exit=1  docs/10-testing.md:490  internal/store/definitely-not-here.go
dead make target   exit=1  docs/10-testing.md:490  make no-such-target-at-all
unread env var     exit=1  docs/10-testing.md:490  FABRIC_TOTALLY_UNREAD_NAME
wrong-case path    exit=1  docs/10-testing.md:490  scripts/Check_Doc_Drift.py
```

### Gates

- 39 tests pass (was 33) · full suite **1829 passed, 1 skipped**
- `check_doc_drift.py` **98.24%** (was 98.05%)
- `ruff check .` and `ty check` clean · `make check` green end to end
- 0.21s over the tree

One doc sentence added to docs/10 recording why existence is asked of git rather than the filesystem — a guard that disagrees with itself by platform is worse than one that is merely wrong.

calvinchengx:
### Iteration 2 review — all four items addressed

| Item | Fix | Test that fails without it |
|---|---|---|
| Fence marker **length** (`` ```` `` block quoting a `` ``` `` one) | `_FENCE` captures the full run; closes only on the same character at length ≥ opener (CommonMark) | `test_a_longer_fence_is_not_closed_by_a_shorter_one` |
| `_TARGET` space-in-class matched leading whitespace → phantom target from a `define` block | spelled as first-name-plus-repeats, cannot start with whitespace | `test_an_indented_line_with_a_colon_is_not_a_rule` |
| `env_names_in_code()` credited the checker's own test file | `SELF` excluded, with a control test proving the exclusion did not widen to every `.py` | `test_the_checker_and_its_own_test_do_not_credit_a_name` |
| `examples/*/README.md` never scanned | scope widened to every per-directory README, intersected with `git ls-files` | `test_a_per_directory_readme_is_in_scope`, `test_an_untracked_file_matching_a_glob_is_not_a_doc` |

The make-flag **recall** note is now stated in the code rather than left implied: only the first word after `make` is read, so rejecting a flag abandons the target behind it. Walking past flags needs a list of which ones consume an argument (`-C` does, `-j4` does not), and getting it wrong reports `portal` as a target — the false positive the guard exists to prevent.

**Widening the scope found a fourth stale reference**, which is the argument for widening it: `e2e/dbt-fabric/README.md:53` sent the reader to a parity document under a number it has never carried. Repointed at `docs/parity.md`.

Intersecting with git is not cosmetic — measured before it, the wider globs pulled in **763 files** and reported drift inside a vendored `dompurify` README. Findings about somebody else's documentation are the fastest way to teach a reader to skim past this check.

One thing worth recording: the first draft of the docs/10 paragraph describing this fix **failed the check**, because it quoted the dead `17-parity` name as a code span. The paragraph now says so.

### Verification

- `check_doc_drift --strict` — **91 docs** (was 64), 3 exemptions, no drift, 0.44s
- **45 tests** (was 39); full suite **1835 passed / 1 skipped**
- Coverage `check_doc_drift.py` **98.31%** (177 stmts, 3 missed); repo total 92.25% vs the 92 floor
- `ruff check .` clean; `make check` exit 0
- **Non-vacuity, twice.** Each of the five code changes reverted individually fails **exactly one** test and no others. Separately, a real doc was broken for each of the three classes against the live tree — including a per-directory README for the newly-widened scope — all three fired, then reverted.

PR #501, opened 2026-09-18
Detect stale Go symbols in docs

calvinchengx:
## Summary
- add a precision-first doc drift class for inline Go package/symbol references under internal/, pkg/, and cmd/
- build the package/symbol index from tracked .go files and keep fenced blocks, file paths, release notes, and EXEMPT behavior aligned with the existing checker
- cover dead/live symbols, path preservation, fenced blocks, non-repo names, and exempt forward references

## Validation
- uv run --frozen python scripts/check_doc_drift.py --strict
- uv run --frozen --group test pytest python/tests/test_check_doc_drift.py -q
- make check

Note: bare  is blocked locally because pyenv does not have the repo's requested Python 3.12 installed; the same command passes through uv.


calvinchengx:
Closing in favour of #506, which is iteration 3 of this same nightshift task rebased onto current `main`.

Measured all three iterations against a probe doc on `main`:

| | live `func New` | live generic `func writePage[T any]` | dead symbol | dead package |
|---|---|---|---|---|
| iter1 (#501) | quiet | **reports it dead** | fires | fires |
| iter2 (#502) | quiet | **reports it dead** | fires | fires |
| iter3 (#506) | quiet | quiet | fires | fires |

This iteration anchors its `func` regex on `(` immediately after the name, so every generic function reads as undeclared. This repo has three — `page`, `writePage` and `writePageKeyed` — so any doc naming one would be reported as drift that is not there. #506 handles `[T any]` and pins the case in its own tests.

Two notes on what was *not* wrong here. The 90,000-line diff is an artefact of branching off `conformance-witness` before #500 merged; rebased onto `main` the real change is 2 files. And the red checks are stale — they ran against a base where `docs/route-coverage.json` had not been re-baselined, not against a defect in this work.

The class itself is sound in all three: zero findings and zero false positives across 92 docs, and non-vacuous on an injected dead symbol.

PR #502, opened 2026-09-18
Detect stale Go symbols in docs

calvinchengx:
## Summary
- add a fourth doc-drift class for repo-qualified Go package/symbol references in inline code spans
- index tracked Go package directories plus obvious top-level func/type/var/const declarations from source text
- keep the detector precision-first: skip fenced blocks, non-repo/non-Go names, file paths, methods, and honor EXEMPT forward references

## Validation
- `uv run --frozen --group test pytest python/tests/test_check_doc_drift.py -q`
- `uv run --frozen python scripts/check_doc_drift.py --strict`
- `make check`

Note: direct `python scripts/check_doc_drift.py --strict` is blocked in this worktree because pyenv cannot find the repo's configured Python 3.12; the uv-backed invocation passes.


calvinchengx:
Closing in favour of #506, which is iteration 3 of this same nightshift task rebased onto current `main`.

Measured all three iterations against a probe doc on `main`:

| | live `func New` | live generic `func writePage[T any]` | dead symbol | dead package |
|---|---|---|---|---|
| iter1 (#501) | quiet | **reports it dead** | fires | fires |
| iter2 (#502) | quiet | **reports it dead** | fires | fires |
| iter3 (#506) | quiet | quiet | fires | fires |

This iteration anchors its `func` regex on `(` immediately after the name, so every generic function reads as undeclared. This repo has three — `page`, `writePage` and `writePageKeyed` — so any doc naming one would be reported as drift that is not there. #506 handles `[T any]` and pins the case in its own tests.

Two notes on what was *not* wrong here. The 90,000-line diff is an artefact of branching off `conformance-witness` before #500 merged; rebased onto `main` the real change is 2 files. And the red checks are stale — they ran against a base where `docs/route-coverage.json` had not been re-baselined, not against a defect in this work.

The class itself is sound in all three: zero findings and zero false positives across 92 docs, and non-vacuous on an injected dead symbol.

